PR #309, opened 2015-01-27
Feature/zen 15834

dbouchillon:
Remove previous fix of adding z property.  Rework method of enabling disabling monitoring by default of services.  
1) User manually enables monitoring on server
2) User creates template for enabling/disabling monitoring specific service
3) User creates template for enabling/disabling monitoring all services


cluther:
**Windows Service Monitoring Test Results**

1. Globally monitor all auto-start services.

	Configuration:

	- Set "Start mode of service to monitor" to Auto in WinService template.

	Result:

	- Didn't work. No services were monitored.

2. Globally disable monitoring of one auto-start service.

	Configuration:

	- Start with #1.
	- Copy WinService template to "BITS" template.
	- Set "Start mode of service to monitor" to None in BITS template.

	Result:

	- Couldn't test due to problem with #1.

3. Globally enable monitoring of one auto-start service.

	Configuration

	- Set "Start mode of service to monitor" to None in WinService template.
	- Copy WinService template to "BITS" template.
	- Set "Start mode of service to monitor" to Auto in BITS template.

	Result:

	- Success.

4. Enable monitoring of one service on one device.

	Configuration

	- Use the service component panel to enable monitoring of a service that
	  would otherwise not be monitored.

	Result:

	- Success.

5. Disable monitoring of one service on one device.

	Configuration

	- Use the service component panel to disable monitoring of a service that
	  would otherwise be monitored.

	Result:

	- Success.

6. Globally enable monitoring of one service regardless of start mode.

	Configuration

	- ???: Currently looks like it can't be done because a single start mode
	  to monitor must be chosen.

---

**Other Notes**

I think both the current "Alert if service is NOT in this state" and "Start mode of service to monitor" both need to be multi-select to be able to handle all use cases.

dougsyer:
One thing to test also it to make sure automatic and automatic(delayed start) both work when you set a service to monitor on automatic startup in the template.  It seems to work that way with the legacy windows monitoring zp but It probably I think its worth a test to to be sure.

dougsyer:
also one challenge is dealing with sql services ...they (at least used to, i havent checked with the latest sql) have the instance name concatenated with the service name and that makes the effort of searching out and monitoring those services a pain at scale.  but not a super big deal since we can cron a script to catch these..

dbouchillon:
Hi Doug, the query we use for services does not distinguish between Automatic and Automatic (Delayed Start).  It will return 'Auto' for both so I believe we're covered on that front.

dougsyer:
thanks i figured.

PR #332, opened 2015-02-27
Fix for ZEN-16195

ssamartyn:


ZenPacks/zenoss/Microsoft/Windows/modeler/WinRMPlugin.py
```diff
@@ -144,6 +148,8 @@ def log_error(self, log, device, error):
                             'http://wiki.zenoss.org/ZenPack:Microsoft_Windows#winrm_setup'
         elif isinstance(error, ConnectionRefusedError):
             message = "Connection refused on %s: Verify WinRM setup"
+            evt_message = "Connection refused on %s: Verify WinRM setup" % device.id
```
cluther: It's redundant to include the device id in the event's message or summary since there's already an event field for it.

ZenPacks/zenoss/Microsoft/Windows/modeler/WinRMPlugin.py
```diff
@@ -212,6 +235,8 @@ def collect(self, device, log):
             try:
                 query_results = yield client.do_collect(
                     conn_info, query_map.iterkeys())
+                msg = "connection for %s is established" % device.id
```
cluther: The device id is redundant here too since it'll already be a field in the event.

PR #382, opened 2015-05-07
Fix for ZEN-17583

ssamartyn:


ZenPacks/zenoss/Microsoft/Windows/datasources/EventLogDataSource.py
```diff
@@ -114,12 +114,23 @@ def config_key(cls, datasource, context):
     @classmethod
     def params(cls, datasource, context):
         te = lambda x: datasource.talesEval(x, context)
+        query = ""
+        query_error = True
+
+        if datasource.query.strip() != '$':
```
cluther: I don't understand the point of this conditional. When would the query ever be the literal '$'?

ZenPacks/zenoss/Microsoft/Windows/datasources/EventLogDataSource.py
```diff
@@ -114,12 +114,23 @@ def config_key(cls, datasource, context):
...
+                pass
+
         return dict(
-            eventlog=te(datasource.eventlog), 
-            query=te(' '.join(string_to_lines(datasource.query))),
-            max_age=te(datasource.max_age), 
-            eventid=te(datasource.id)
-        )
+                eventlog=te(datasource.eventlog), 
+                query=query,
```
cluther: If the above TALES evaluation raised an exception, `query` would be an empty string here. That wouldn't be very useful for understanding the event. It would be better to use the unevaluated datasource.query that the user configured if it wasn't able to be TALES evaluated.

ZenPacks/zenoss/Microsoft/Windows/datasources/EventLogDataSource.py
```diff
@@ -127,6 +138,15 @@ def collect(self, config):
         log.info('Start Collection of Events')
 
         ds0 = config.datasources[0]
+
+        try:
```
cluther: What's the purpose of handling this exception you're generating here anyway? Wouldn't the following achieve the same thing?

```python
if ds0.params['query_error']:
    raise EventLogException('Please verify EventQuery on datasource %s'
        % ds0.params['eventid'])
```

PR #384, opened 2015-05-07
Fixes ZEN-17683

dbouchillon:
Need to have backwards compatibility with zDBInstances.  We should also allow for users to enter different credentials for different instances.  Fixed default mssqlserver instance also.


ZenPacks/zenoss/Microsoft/Windows/datasources/ShellDataSource.py
```diff
@@ -713,6 +648,7 @@ def collect(self, config):
         counters = [dsconf.params['resource'] for dsconf in config.datasources]
 
         if dsconf0.params['strategy'] == 'powershell MSSQL':
+            import pdb; pdb.set_trace()
```
cluther: PDB!

PR #394, opened 2015-05-28
JIRA 18049

cajoz:
* changed default value of `points` class attribute on WinCmd to None
* added initializer method for WinCmd to set `points` instance attribute to empty list
* cut worse-than-useless rethrow


cluther:
Opened #395 with the same patch that's in this pull request to have it applied to the hotfix/2.4.4 branch.

PR #411, opened 2015-07-28
Fixes ZEN-18947

dbouchillon:
for better performance, we should only run this query once to get all service statuses then parse for monitored services in onSuccess.

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/datasources/ServiceDataSource.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/datasources/ServiceDataSource.py
@@ -163,9 +163,6 @@ class ServicePlugin(PythonDataSourcePlugin):
             datasource.getCycleTime(context),
             datasource.id,
             datasource.plugin_classname,
-            params.get('servicename'),
-            params.get('alertifnot'),
-            params.get('startmode'),
         )
 
     @classmethod
@@ -181,8 +178,6 @@ class ServicePlugin(PythonDataSourcePlugin):
         params['startmode'] = datasource.talesEval(
             datasource.startmode, context)
 
-
-
         return params
 
     @defer.inlineCallbacks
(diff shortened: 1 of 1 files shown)


ZenPacks/zenoss/Microsoft/Windows/datasources/ServiceDataSource.py
```diff
@@ -208,50 +201,76 @@ def collect(self, config):
...
-            })
+                                'eventClass': "/Status",
+                                'severity': ZenEventClasses.Warning,
+                                'eventClassKey': 'WindowsServiceCollectionError',
+                                'eventKey': 'WindowsServiceCollection',
+                                'summary': 'No results returned for service query',
+                                'device': config.id})
+            return data
+
+        for index in range(0, len(serviceinfo)):
```
cluther: You could just use range(len(serviceinfo)) here. The range function defaults to starting from 0 if only one argument is provided.

ZenPacks/zenoss/Microsoft/Windows/datasources/ServiceDataSource.py
```diff
@@ -208,50 +201,76 @@ def collect(self, config):
...
-                'eventClass': eventClass,
-                'eventClassKey': 'WindowsServiceLog',
-                'eventKey': eventKey,
-                'severity': ZenEventClasses.Clear,
-                'summary': evtmsg,
-                'component': prepId(serviceinfo[0].Name),
-                'device': config.id,
-            })
+                                'eventClass': "/Status",
+                                'severity': ZenEventClasses.Warning,
```
cluther: This seems more like an Error severity than Warning. As would the same down in the onError method.

ZenPacks/zenoss/Microsoft/Windows/datasources/ServiceDataSource.py
```diff
@@ -208,50 +201,76 @@ def collect(self, config):
...
+            eventKey = service['eventKey'] if service['eventKey'] else "WindowsService"
+
+            if serviceinfo[index].State != service['alertifnot']:
+
+                evtmsg = 'Service Alert: {0} has changed to {1} state'.format(
+                    serviceinfo[index].Name,
+                    serviceinfo[index].State
+                )
+
+                data['events'].append({
```
cluther: It would be useful to add additional details from serviceinfo to the event such as unaltered values for Name and State. These things are useful in transforms so you don't have to use a regex to try to parse them out of the summary.

PR #415, opened 2015-08-12
Fixes dcdiag problems

dbouchillon:
Specify username and password on the command line to fix issue of access denied.  Also we only need to run dcdiag every 30 minutes instead of 5 minutes.

Add in latest version of txwinrm


ZenPacks/zenoss/Microsoft/Windows/datasources/ShellDataSource.py
```diff
@@ -203,9 +203,13 @@ class DCDiagStrategy(object):
...
 
-    def build_command_line(self, tests, testparms):
+    def build_command_line(self, tests, testparms, username, password):
         self.run_tests = set(tests)
-        dcdiagcommand = 'dcdiag /q /test:' + ' /test:'.join(tests)
+        try:
+            dcuser = '{}\\{}'.format('.'.join(username.split('@')[1].split('.')[:-1]),username.split('@')[0])
+        except Exception:
+            raise WindowsShellException('Username invalid.  Must be in user@example.com format.  Check zWinRMUser: {}'.format(username))
+        dcdiagcommand = 'dcdiag /q /u:{} /p:{} /test:'.format(dcuser, password) + ' /test:'.join(tests)
```
cluther: How necessary is it to pass credentials on the command line? This seems like a pretty serious security issue for anyone who can see the process table on the system.

PR #418, opened 2015-08-13
Fix for 18770

ssamartyn:


ZenPacks/zenoss/Microsoft/Windows/datasources/IISSiteDataSource.py
```diff
@@ -132,10 +135,10 @@ def collect(self, config):
         wql_iis7 = 'select ServerAutoStart from Site where name="{0}"'.format(
             ds0.params['statusname'])
 
-        WinRMQueries = [
-            create_enum_info(wql=wql_iis6, resource_uri=resource_uri_iis6),
-            create_enum_info(wql=wql_iis7, resource_uri=resource_uri_iis7),
-            ]
+        if ds0.params['iis_version'] == 6:
+            WinRMQueries = [create_enum_info(wql=wql_iis6, resource_uri=resource_uri_iis6),]
+        else:
```
cluther: This is going to run the IIS7 query if ds0.params['iis_version'] is None, which it definitely will be before the device is remodeled with the new modeler plugin. You should probably raise a warning event in the case that you don't know which query to run. Otherwise you'll probably end up with another event that offers no clue as to what the real problem is.

PR #431, opened 2015-08-25
ZEN-18836 Microsoft Windows - Database event is in Unknown class

byshovets:
Fixes https://jira.zenoss.com/browse/ZEN-18836


ZenPacks/zenoss/Microsoft/Windows/datasources/ShellDataSource.py
```diff
@@ -970,8 +970,9 @@ def onSuccess(self, results, config):
                         prepare_instance(dsconf.component),
                         value[1].strip()
                     )
-
+                eventClass = dsconf.eventClass if dsconf.eventClass else "/Status"
```
ssvsergeyev: It may works in shorter way
eventClass = dsconf.eventClass or "/Status"

PR #432, opened 2015-08-25
ZEN-18835: Microsoft Windows - Cluster event is in Unknown class

byshovets:
Fixes https://jira.zenoss.com/browse/ZEN-18835


ZenPacks/zenoss/Microsoft/Windows/datasources/ShellDataSource.py
```diff
@@ -1010,7 +1010,9 @@ def onSuccess(self, results, config):
                             'Failed': ZenEventClasses.Critical
                         }.get(value[1], ZenEventClasses.Info)
 
+                        eventClass = dsconf.eventClass or "/Status"
```
ssvsergeyev: variable used 1 time, and same as in 431 may work here

PR #435, opened 2015-08-26
Fix for ZEN-18823

ssamartyn:


ZenPacks/zenoss/Microsoft/Windows/datasources/ProcessDataSource.py
```diff
@@ -418,29 +418,32 @@ def onSuccess(self, results, config):
...
+                            "%s %s %s: Couldn't convert %r to integer",
+                            datasource.device, datasource.component, point.id,
+                            value)
+                    except AttributeError:
+                        LOG.warn(
+                            "%s %s: %s not in result",
+                            datasource.device, datasource.component, point.id)
+                    else:
+                        metrics_by_component[datasource.component][point.id].append(value)
+        except Exception:
```
cluther: This exception handling is too coarse. What actually causes the IndexError described in ZEN-18823? If it's the creation of `perf_key` using the `[0]` of that list comprehension, maybe we should build that list first and just check that it's len() > 0 before doing the rest.

It's also not clear to me what situation actually causes that list comprehension to return an empty list. Is it really nothing to worry about, or is it a sign that that we were unable to get data? If we were unable to get data we expected to get, simply passing along silently doesn't seem like a good idea.

ssamartyn: Chet, I was not able to reproduce this issue and for now have no idea what is the reason why part of result is empty/broken. Looks like this happens on heavy loaded and we get not fully prepared results from server.
So we made how was mentioned in issue expected results : " If the list is empty, don't reference a specific location in the list."

PR #441, opened 2015-09-03
Fix for ZEN-18638

ssamartyn:


dbouchillon:
The Windows Service panel has been deprecrated.  We now use the WinService datasource instead.

PR #502, opened 2015-11-20
Fixes ZEN-21029

dbouchillon:
https://jira.zenoss.com/browse/ZEN-21029

Version 7 and up of IIS uses same namespace and wmi class.  Better to check the version in the ZenPack.  This also helps with debugging


ZenPacks/zenoss/Microsoft/Windows/datasources/IISSiteDataSource.py
```diff
@@ -162,7 +162,16 @@ def collect(self, config):
         if not iis_version:
             winrs = IISCommander(conn_info)
             version = yield winrs.get_iis_version()
-            iis_version = version.stdout[0]
+            # version should be in 'Version x.x' format
+            # 7 and above use the same namespace/query
+            try:
+                iis_version = re.match('Version (\d).*', version.stdout[0]).group(1)
```
cluther: This seems like it could also raise an AttributeError if `re.match()` doesn't match because you'll be attempting to call `None.group(1)`.

PR #511, opened 2015-12-16
Fixes ZEN-21325

dbouchillon:
Similar fix as in https://jira.zenoss.com/browse/ZEN-21242.  Problem is that not all cluster disks are local to the node and some information may not be found when looping through results of call to get-clusterresource.  Also moved $diskinfo outside of the loop as this only needs to be done once.


eimajenthat:
Looked at this with @dbouchillon because it's bizarre and strange.  Here's the handy dandy gist:  https://gist.github.com/dbouchillon/498510177553f0a88847

PR #519, opened 2016-02-01
Fix for Zendesk 104993

ssamartyn:


ZenPacks/zenoss/Microsoft/Windows/datasources/ShellDataSource.py
```diff
@@ -318,8 +318,11 @@ def parse_result(self, config, result):
         # Give error feedback to user
         eventClass = dsconf.eventClass if dsconf.eventClass else "/Status"
         if result.stderr:
-            err_index = [i for i, val in enumerate(result.stderr) if "At line:" in val][0]
-            msg = 'Custom Command error: ' + ''.join(result.stderr[:err_index])
+            try:
+                err_index = [i for i, val in enumerate(result.stderr) if "At line:" in val][0]
+                msg = 'Custom Command error: ' + ''.join(result.stderr[:err_index])
+            except:
```
cluther: Don't use a bare `except`. In cases like these use `except Exception` at the very least to avoid potentially capturing SystemExit or KeyboardInterrupt.

PR #527, opened 2016-03-14
Link points to deleted page

aseques:


dbouchillon:
Hi, thank you for your interest in the Microsoft Windows ZenPack.  Unfortunately our policy is that we do not accept external contribution to the ZenPack.

Thanks

PR #529, opened 2016-03-21
Fixes ZEN-22558

dbouchillon:
Chet please review

We need to do our part to kill dead connections so we remove open files.  A few pep8 updates.  Update datasources to make sure to use @defer.inlineCallbacks


dbouchillon:
wrong fix for the situation

PR #536, opened 2016-03-29
fix possible IndexError

dbouchillon:
Fixes ZEN-22369

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/modeler/plugins/zenoss/winrm/OperatingSystem.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/modeler/plugins/zenoss/winrm/OperatingSystem.py
@@ -57,7 +57,7 @@ class OperatingSystem(WinRMPlugin):
         operatingSystem = results.get('Win32_OperatingSystem', (None,))[0]
         clusterInformation = results.get('MSCluster', ())
         exchange_version = results.get('exchange_version')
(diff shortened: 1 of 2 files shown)


ZenPacks/zenoss/Microsoft/Windows/modeler/plugins/zenoss/winrm/OperatingSystem.py
```diff
@@ -97,7 +97,7 @@ def process(self, device, results, log):
             pass
 
         # if NTDS service present then this is a DC
-        if domainController:
+        if len(domainController):
```
joshw: Technically don't really need to change this..  if domainController would do the same thing, i believe.

ssOleg: len((None,)) == True
We get true in all cases.

byshovets: So line 60 can be changed to: domainController = results.get('ActiveDirectory', ())

dbouchillon: if there are no results returned, domainController results in an empty list, []
so len([]) == 0

i made another change to make this less confusing

PR #537, opened 2016-03-30
Look for string in message, not result instance

dbouchillon:
Fixes ZEN-20232

result is not iterable.  we should look at the actual message
we probably won't see this message anymore, but just in case
it slips by we'll keep it.


dbouchillon:
bump.  super simple review

ZenPacks/zenoss/Microsoft/Windows/datasources/EventLogDataSource.py
```diff
@@ -325,7 +325,7 @@ def onError(self, result, config):
         if isinstance(result.value, (MissedEventLogException, InvalidEventQueryValue)):
             msg = "WindowsEventLog: " + result.value.message
         severity = ZenEventClasses.Warning
-        if 'This cmdlet requires Microsoft .NET Framework version 3.5 or greater' in result:
+        if 'This cmdlet requires Microsoft .NET Framework version 3.5 or greater' in result.value.message:
```
ssOleg: Probably it is better to replace 'result.value.message' to 'msg'

dbouchillon:
changed to 'msg'

PR #541, opened 2016-04-01
Update txwinrm lib

dbouchillon:
Fixes ZEN-22558


dbouchillon:
change was made in txwinrm and merged already

https://github.com/zenoss/txwinrm/pull/77

PR #543, opened 2016-04-18
Added @save decorator for help in diagnosing customer issues

dbouchillon:
@save will dump out arguments to functions.  Users can then zip them up
and send to us so we can work with their data.


dbouchillon:
bump

PR #558, opened 2016-05-09
Change WMI to WinRM.

dbouchillon:
Fixes ZEN-23115

Not sure why this was ever set to WMI since this is the WinRM ZenPack


dbouchillon:
gentle nudge

PR #559, opened 2016-05-10
Fixes ZEN-23156

j053ph4:
- changed "query" in _properties from "lines" to "string", as it is
treated like a string throughout datasource
- added migrate script to correct any EventLogDatasources with list
query
- found and converted numerous windows newline-delimited files to unix
newline-delimited


dbouchillon:
we'll lose git history of code by changing windows newlines to unix newlines.  we can't do that

also, we need to leave query as a 'lines' type because of the datasource dialog

PR #569, opened 2016-05-19
Fixes ZEN-22640

j053ph4:
- Rewrote Device getRRDTemplates method to abstract the various software
detection criteria to their own methods, where they can be fine-tuned
further if needed.

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/Device.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/Device.py
@@ -131,10 +131,6 @@ class Device(BaseDevice):
 
     def is_ntds(self):
         '''Return True if an NTDS/AD server'''
-        # running the NTDS service indicates that this is an AD server
-        for component in self.getDeviceComponents():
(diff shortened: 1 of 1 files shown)


ZenPacks/zenoss/Microsoft/Windows/Device.py
```diff
@@ -121,46 +121,56 @@ def getClusterMachinesList(self):
...
+        # if we have IIS components, then we are IIS server
+        for component in self.getDeviceComponents():
+            if isinstance(component, WinIIS):
+                return True
+        return False
+
+    def is_ntds(self):
+        '''Return True if an NTDS/AD server'''
+        # running the NTDS service indicates that this is an AD server
+        for component in self.getDeviceComponents():
```
dbouchillon: can you remove this check?  apparently checking for ntds is not the best way as a server can have lightweight ad on it and not be a domain controller.  I added a fix in the os modeler plugin recently that is the best way to detect a dc so we should only look at the self.domain_controller property.

PR #571, opened 2016-05-19
Make winrm ping more understandable

dbouchillon:
Fixes ZEN-22166

Updated the readme to explain winrm ping and why it's better than a
simple ping.  Also, we'll add any error message to the device down summary
so users will know why it's down.


dbouchillon:
bump

dbouchillon:
bump

PR #578, opened 2016-06-03
Potential Fix for ZEN-20148

j053ph4:
CHANGES
- rebase WinService class on Products.ZenModel.WinService so that it
inherits serviceclass relation
- added additional methods to "monitored" so that it takes Service class
into account
- updated modeler script to reuse "winservices" relation
- added migrate script to rebuild WinService objects

MISC
- added __init__.py to migrate so that scripts will run
- fixed ZenPack class issue in base __init__.py


ZenPacks/zenoss/Microsoft/Windows/WinService.py
```diff
@@ -108,6 +109,48 @@ def monitored(self):
...
 
+        # 3 check the service class
+        if hasattr(self, 'serviceclass') and 'serviceclass' in self.getRelationshipNames():
+            sc = self.serviceclass()
+            if sc:
+                org = sc.serviceorganizer()
+                # check the service class monitored start modes
+                if self.startMode in sc.monitoredStartModes:
+                    # check the zMonitor organizer property
+                    if org and hasattr(org, 'zMontitor'):
```
dbouchillon: should it be `zMonitor`?

j053ph4: I believe so, it's from ZenModel/ServiceOrganizer.py and the Service Class page has:

Enable Monitoring? (zMonitor)

dbouchillon: sorry.  just meant to say that it was misspelled

j053ph4: Oops you're correct.  For some reason I mis-spell "monitor" all of the time and can't see it.

PR #581, opened 2016-06-03
Fixes ZEN-23326

j053ph4:
- Python twisted Failure object is different in 5.x, so rewrote to
account for differences.


j053ph4:
bump

PR #583, opened 2016-06-17
Enable the windows services page

dbouchillon:
Fixes ZEN-20148

Subclass platform WinService so services will be shown on services
page.  Existing winrmservices will be removed on a remodel.  Users
can enable monitoring on services page now.  Remove deprecated
message when navigating to windows services page

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/WinService.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/WinService.py
@@ -110,6 +110,7 @@ class WinService(BaseWinService):
                         return True
 
         # 3 check the service class
+        # be sure we can get the serviceclass and that we have a relationship with serviceclass
         if hasattr(self, 'serviceclass') and 'serviceclass' in self.getRelationshipNames():
             sc = self.serviceclass()
             if sc:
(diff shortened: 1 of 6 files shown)


dbouchillon:
bump

ZenPacks/zenoss/Microsoft/Windows/WinService.py
```diff
@@ -109,6 +109,74 @@ def monitored(self):
                     if self.getMonitored(datasource):
                         return True
 
+        # 3 check the service class
+        if hasattr(self, 'serviceclass') and 'serviceclass' in self.getRelationshipNames():
```
joshw: I don't think I fully understand what is going on here.   I'm guessing, though unsure, that this has something to do with existing objects in ZODB that were created before you changed the base class of this class?

It feels to me like they should be migrated over in a migrate script so that everything is sane, but I also might be missing something.   

So I think at a minimum, a comment is needed here explaining it, but possibly a migration script.

dbouchillon: When the Microsoft.Windows ZenPack was made, it was decided to move the monitoring of services to monitoring templates instead of using the "serviceclass" or Infrastructure -> Windows Services page.  We are now enabling the page so that services can be organized together for viewing status.  This will also allow the user to enable monitoring from that page.  It is a similar thing to ip services and os processes.

joshw: That's fine, it is more that funny dance of checking for the existance of the attribute and relationship that is confusing to me.

PR #589, opened 2016-06-21
Fixes ZEN-23648

j053ph4:
- cleaned up details pane for Interface and FileSystem components


ZenPacks/zenoss/Microsoft/Windows/zenpack.yaml
```diff
@@ -415,6 +418,11 @@ classes:
     dynamicview_relations:
       impacts: []
       impacted_by: [device]
+    relationships:
+      DEFAULTS:
+        grid_display: true
+        details_display: false
+      teaminterface: {}
```
dbouchillon: This relationship needs to be there

PR #590, opened 2016-06-21
Fixes ZEN-23713

j053ph4:
- override WinService's getFailSeverity method to take into account
template-based  as well as service class severities
- revise ServiceDataSource to use WinService's getFailSeverity
- revise info.py and interface.py to include Fail Severity drop-down
(subclass WinServiceInfo/IWinServiceInfo)
- preserve info.py/interface.py overrides for usermonitor property
- revise javascript to better align with legacy while providing
formatting enhancements (column width)

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/WinService.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/WinService.py
@@ -96,17 +96,9 @@ class WinService(BaseWinService):
             return self.monitor
 
         # 2 - Check what our template says to do.
-        template = self.getRRDTemplate()
-        if template:
-            datasource = template.datasources._getOb('DefaultService', None)
-            if datasource:
-                if self.getMonitored(datasource):
-                    return True
-            # 3 - Allow for other datasources to be specified.
-            for datasource in template.getRRDDataSources():
-                if datasource.id != 'DefaultService' and hasattr(datasource, 'startmode'):
-                    if self.getMonitored(datasource):
-                        return True
+        datasource = self.getMonitoredDataSource()
+        if datasource and self.getMonitored(datasource):
+            return True
 
         # 3 check the service class
         # be sure we can get the serviceclass and that we have a relationship with serviceclass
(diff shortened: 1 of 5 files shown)


ZenPacks/zenoss/Microsoft/Windows/WinService.py
```diff
@@ -164,10 +164,33 @@ def get_winservices_modes(self):
...
+                                'monitor': svc.monitored(),
+                                'severity': svc.getFailSeverity(),
                                 }
         return data
 
+    def getTemplateDefaultService(self):
+        '''Return datasource for template if it exists'''
+        template = self.getRRDTemplate()
+        if template:
+            return template.datasources._getOb('DefaultService', None)
```
dbouchillon: Users can also create a new datasource for monitoring services, so it won't always be the 'DefaultService'.  you'll probably need to loop through them as we are in monitored() and call getMonitored to see if the datasource applies.

Or, what might be better is to add a new attribute to WinService to save the datasource name.  you can set it when a datasource enables monitoring.  it can be empty when there is no datasource

ZenPacks/zenoss/Microsoft/Windows/resources/js/global.js
```diff
@@ -80,7 +87,7 @@ Ext.define("Zenoss.component.WinRMServicePanel", {
     }
 });
 
-ZC.registerName('WinRMService', _t('Service'), _t('Services'));
```
dbouchillon: instead of making changes to this js, let's just use the existing one in platform.  doing this stuff makes it look just like that one anyway

ZenPacks/zenoss/Microsoft/Windows/WinService.py
dbouchillon: Also, the service will always have a template associated with it.  You should instead save the datasource name as an attribute and return the severity on it

something like this:
before line 98:
`self.datasource_name = ''`

before returning True on lines 104, 109, then set the name
`self.datasource_name = datasource.id`

then your new method would be 
```
getTemplateDataSource(self):
    if self.datasource_name:
        template = self.getRRDTemplate()
        return template.datasources._getOb(self.datasource_name, None)
```

ZenPacks/zenoss/Microsoft/Windows/resources/js/global.js
j053ph4: I left the .js as-is, since preexisting WinService components will still have their metatype as "WinRMService" prior to the modeler running after upgrade of the ZP.  Not a huge deal, but thought it might alarm users to see that.

PR #591, opened 2016-06-21
change subclass for service datasource info

dbouchillon:
Fixes ZEN-19615

Subclass interface and info off of base and add in only what we need
Change event class to /Status/WinService
We should not allow users to select an event class for service monitoring.
We need to see status by looking in /Status/WinService.


dbouchillon:
bump

PR #593, opened 2016-06-22
Fixes ZEN-23713

j053ph4:
- override WinService's getFailSeverity method to take into account
template-based  as well as service class severities
- revise ServiceDataSource to use WinService's getFailSeverity
- revise info.py and interface.py to include Fail Severity drop-down
(subclass WinServiceInfo/IWinServiceInfo)
- preserve info.py/interface.py overrides for usermonitor property
- revise javascript to better align with legacy while providing
formatting enhancements (column width)
- Revised template datasource identification method

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/WinService.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/WinService.py
@@ -202,7 +202,8 @@ class WinService(BaseWinService):
                 # if it's still undefined, check for other datasources
                 if not self.datasource_id:
                     for datasource in template.getRRDDataSources():
-                        if datasource.id == 'DefaultService': continue
+                        if datasource.id == 'DefaultService': 
+                            continue
                         test_datasource(datasource)
 
     def getFailSeverity(self):
@@ -212,8 +213,8 @@ class WinService(BaseWinService):
         template = self.getRRDTemplate()
         if template:
             datasource = template.datasources._getOb(self.datasource_id, None)
-        if datasource:
-            return datasource.severity
+            if datasource:
+                return datasource.severity
         return self.getAqProperty("zFailSeverity")
 
     def getFailSeverityString(self):
--- a/ZenPacks/zenoss/Microsoft/Windows/WinService.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/WinService.py
@@ -22,7 +22,6 @@ class WinService(BaseWinService):
     '''
     Model class for Windows Service.
     '''
-    meta_type = portal_type = 'WinRMService'
 
     servicename = None
     caption = None
@@ -31,6 +30,7 @@ class WinService(BaseWinService):
     account = None
     monitor = False
     usermonitor = False
+    datasource_id = None
 
     _properties = BaseWinService._properties + (
         {'id': 'servicename', 'label': 'Service Name', 'type': 'string'},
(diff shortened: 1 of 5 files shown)


ZenPacks/zenoss/Microsoft/Windows/WinService.py
```diff
@@ -164,10 +155,35 @@ def get_winservices_modes(self):
...
+                                'monitor': svc.monitored(),
+                                'severity': svc.getFailSeverity(),
                                 }
         return data
 
+    def getMonitoredDataSource(self):
+        '''Return datasource for template if it exists'''
+        template = self.getRRDTemplate()
+        if template:
+            for datasource in template.getRRDDataSources():
```
dbouchillon: Because of the order of precedence that we've outlined, you'll need to check specifically for the 'DefaultService' first, then loop through the rest if necessary.

ZenPacks/zenoss/Microsoft/Windows/WinService.py
```diff
@@ -97,17 +96,9 @@ def monitored(self):
...
-            datasource = template.datasources._getOb('DefaultService', None)
-            if datasource:
-                if self.getMonitored(datasource):
-                    return True
-            # 3 - Allow for other datasources to be specified.
-            for datasource in template.getRRDDataSources():
-                if datasource.id != 'DefaultService' and hasattr(datasource, 'startmode'):
-                    if self.getMonitored(datasource):
-                        return True
+        datasource = self.getMonitoredDataSource()
```
dbouchillon: This should be an instance variable so we can reference it later in getFailSeverity().  That's going to be a lot of looping on each cycle for each and every service that's monitored.  A single device has ~120 services.  say half of those are monitored on each device and you have several hundred devices.  that's a lot of unnecessary computations.

ZenPacks/zenoss/Microsoft/Windows/WinService.py
```diff
@@ -164,10 +157,50 @@ def get_winservices_modes(self):
...
                                 }
         return data
 
+    def getMonitoredDataSource(self):
+        '''Return datasource for template if it exists'''
+        
+        template = self.getRRDTemplate()
+        if template:
+            # look for cached datasource reference
+            if self.datasource_id is not None:
```
dbouchillon: Once datasource_id is set, it will never get past this if there is a change.

ZenPacks/zenoss/Microsoft/Windows/WinService.py
```diff
@@ -164,10 +157,50 @@ def get_winservices_modes(self):
...
+                    return datasource
+        # if not successful set back to None
+        self.datasource_id = None
+        return None
+
+    def getFailSeverity(self):
+        """
+        Return the severity for this service when it fails.
+        """
+        datasource = self.getMonitoredDataSource()
```
dbouchillon: here you could get the template, then get the datasource object using self.datasrouce_id

ZenPacks/zenoss/Microsoft/Windows/WinService.py
```diff
@@ -40,6 +39,8 @@ class WinService(BaseWinService):
         {'id': 'account', 'label': 'Account', 'type': 'string'},
         {'id': 'usermonitor', 'label': 'User Selected Monitor State',
             'type': 'boolean'},
+        {'id': 'datasource_id', 'label': 'Account', 'type': 'string'},
```
dbouchillon: datasource_id doesn't need to be a property.  it's just an internal attribute

ZenPacks/zenoss/Microsoft/Windows/WinService.py
```diff
@@ -161,13 +157,71 @@ def get_winservices_modes(self):
...
+                        test_datasource(datasource)
+
+    def getFailSeverity(self):
+        """
+        Return the severity for this service when it fails.
+        """
+        template = self.getRRDTemplate()
+        if template:
+            datasource = template.datasources._getOb(self.datasource_id, None)
+        if datasource:
```
dbouchillon: probably wouldn't happen, but if template wasn't found, then datasource would not be set and we'd get an exception.  probably best to just indent these 2 lines

ZenPacks/zenoss/Microsoft/Windows/WinService.py
```diff
@@ -161,13 +157,71 @@ def get_winservices_modes(self):
...
+            template = self.getRRDTemplate()
+            if template:
+                # first check DefaultService
+                datasource = template.datasources._getOb('DefaultService', None)
+                if datasource:
+                    test_datasource(datasource)
+                # if it's still undefined, check for other datasources
+                if not self.datasource_id:
+                    for datasource in template.getRRDDataSources():
+                        if datasource.id == 'DefaultService': continue
```
dbouchillon: pep8 doesn't like this.  can you move continue to next line?

PR #596, opened 2016-06-23
Add description back in, check for datasource.enabled

dbouchillon:
Fixes ZEN-23761

Was not checking for datasource.enabled attr
Added description back into the details pane

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/WinService.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/WinService.py
@@ -33,11 +33,7 @@ class WinService(BaseWinService):
     datasource_id = None
 
     _properties = BaseWinService._properties + (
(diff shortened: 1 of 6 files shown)


ZenPacks/zenoss/Microsoft/Windows/WinService.py
```diff
@@ -98,7 +98,7 @@ def monitored(self):
 
         # 2 - Check what our template says to do.
         datasource = self.getMonitoredDataSource()
-        if datasource and self.getMonitored(datasource):
+        if datasource and self.getMonitored(datasource) and datasource.enabled:
```
j053ph4: Does this change also need to be made to "getMonitoredDataSource" (line 179) and "setMonitoredDataSource" (line 191)?

PR #612, opened 2016-06-30
Fixes ZEN-23838

j053ph4:
- rebuilt analytics_bundle.zip


dbouchillon:
Looks fine to me.  The error she's seeing is a problem in the analytics server.  the id is not in the analytics bundle

PR #614, opened 2016-07-01
Fixes ZEN-23848

j053ph4:
- reportable.zcml was inadvertently commented out during testing
- rebuilt analytics_bundle and schema.data files


dbouchillon:
The Microsoft Windows Domain still contains Cluster items Disk, Nodes, etc.  Those need to be removed from your device

dbouchillon:
Line 381 in windows schema, change Device name to Service Name

PR #620, opened 2016-07-26
Service monitoring should follow the order of precedence

dbouchillon:
Fixes ZEN-24165

Enabling/disabling monitoring for a service should obey the 4 laws
Simplified the monitored() method.  We can get all the
information we need from the context when a datasource is created.
We know which services are monitored because there are datasources
for the ones that are being monitored.  We should also reindex
services if a datasource is removed.

update readme

fix version #


dbouchillon:
bump

PR #630, opened 2016-08-24
Make sure we're finding jobs correctly.

dbouchillon:
Fixes ZEN-24519

Previously we were assuming that 'master' was the db at pos [0].
Instead this will use the JobServer property of the sql server.

update readme


dbouchillon:
bump

PR #631, opened 2016-09-02
Respect Component WinService template over all others

dbouchillon:
Fixes ZEN-24528

If a user has copied the default template in order to disable monitoring
we should honor that before any other datasources in the default template


dbouchillon:
bump

PR #633, opened 2016-09-20
Use new SessionManager in txwinrm to send requests

dbouchillon:
Fixes ZEN-24477

* Add in new txwinrm library
* Change plugins to use new clients to send/receive requests


dbouchillon:
bump

dbouchillon:
Bump

PR #644, opened 2016-11-02
Added additional event classes to cover connection errors.

itshane:
Fixes ZEN-25700

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/objects/objects.xml
+++ b/ZenPacks/zenoss/Microsoft/Windows/objects/objects.xml
@@ -16782,9 +16782,34 @@ True
 </object>
 <!-- ('', 'zport', 'dmd', 'Events', 'Status', 'Winrm') -->
 <object id='/zport/dmd/Events/Status/Winrm' module='Products.ZenEvents.EventClass' class='EventClass'>
+<object id='Ping' module='Products.ZenEvents.EventClass' class='EventClass'>
 </object>
-<!-- ('', 'zport', 'dmd', 'Events', 'Status', 'Winrm', 'Ping') -->
-<object id='/zport/dmd/Events/Status/Winrm/Ping' module='Products.ZenEvents.EventClass' class='EventClass'>
+<object id='Auth' module='Products.ZenEvents.EventClass' class='EventClass'>
+<object id='PasswordExpired' module='Products.ZenEvents.EventClass' class='EventClass'>
+<tomanycont id='instances'>
+<object id='Password Expired Default' module='Products.ZenEvents.EventClassInst' class='EventClassInst'>
+<property type="string" id="eventClassKey" mode="w" >
+MW|PasswordExpired
+</property>
+<property type="int" id="sequence" mode="w" >
+1001
+</property>
+</object>
+</tomanycont>
+</object>
+<object id='WrongCredentials' module='Products.ZenEvents.EventClass' class='EventClass'>
+<tomanycont id='instances'>
+<object id='Wrong Credentials Default' module='Products.ZenEvents.EventClassInst' class='EventClassInst'>
+<property type="string" id="eventClassKey" mode="w" >
+MW|WrongCredentials
+</property>
+<property type="int" id="sequence" mode="w" >
+1002
+</property>
+</object>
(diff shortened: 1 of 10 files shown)


ZenPacks/zenoss/Microsoft/Windows/utils.py
```diff
@@ -362,11 +362,18 @@ def dumper(self, *args, **kwargs):
...
 
-def checkExpiredPassword(config, events, error):
-    '''add password expired event'''
-    if 'Password expired' in error or 'Check username and password' in error:
+def errorMsgCheck(config, events, error):
+    """Check error message and generate appropriate event."""
+    if 'Password expired' in error:
         events.append({
-            'eventClass': '/Status/Winrm/Ping',
+            'eventClass': '/Status/Winrm/Auth/PasswordExpired',
```
cluther: It's preferable to set an event's eventClassKey instead of eventClass when creating events. This allows the user far better control over the events if they're not satisfied with the default event class, severity, etc.

I would recommend setting eventClassKey appropriately here, and adding default event class mappings to bring them into the desired event classes by default.

ZenPacks/zenoss/Microsoft/Windows/objects/objects.xml
```diff
@@ -16782,9 +16782,40 @@ True
...
 <!-- ('', 'zport', 'dmd', 'Events', 'Status', 'Winrm') -->
 <object id='/zport/dmd/Events/Status/Winrm' module='Products.ZenEvents.EventClass' class='EventClass'>
+<object id='Ping' module='Products.ZenEvents.EventClass' class='EventClass'>
+</object>
+<object id='Auth' module='Products.ZenEvents.EventClass' class='EventClass'>
+<object id='PasswordExpired' module='Products.ZenEvents.EventClass' class='EventClass'>
+<tomanycont id='instances'>
+<object id='Password Expired Default' module='Products.ZenEvents.EventClassInst' class='EventClassInst'>
+<property type="string" id="eventClassKey" mode="w" >
+defaultmapping
```
cluther: This should be _MW|PasswordExpired_ instead of _defaultmapping_. Using defaultmapping is more costly when processing events, and should only be used when necessary.

ZenPacks/zenoss/Microsoft/Windows/objects/objects.xml
```diff
@@ -16782,9 +16782,40 @@ True
...
+</property>
+<property type="int" id="sequence" mode="w" >
+1001
+</property>
+<property type="string" id="rule" mode="w" >
+(getattr(evt, 'eventClassKey', '') or '').startswith('MW|PasswordExpired')
```
cluther: This _rule_ can be removed once the mapping's eventClassKey is changed to _MW|PasswordExpired_.

ZenPacks/zenoss/Microsoft/Windows/objects/objects.xml
```diff
@@ -16782,9 +16782,40 @@ True
...
+</property>
+</object>
+</tomanycont>
+</object>
+<object id='WrongCredentials' module='Products.ZenEvents.EventClass' class='EventClass'>
+<tomanycont id='instances'>
+<object id='Wrong Credentials Default' module='Products.ZenEvents.EventClassInst' class='EventClassInst'>
+<property type="string" id="eventClassKey" mode="w" >
+defaultmapping
```
cluther: This should be _MW|WrongCredentials_ instead of _defaultmapping_. Using defaultmapping is more costly when processing events, and should only be used when necessary.

ZenPacks/zenoss/Microsoft/Windows/objects/objects.xml
```diff
@@ -16782,9 +16782,40 @@ True
...
+</property>
+<property type="int" id="sequence" mode="w" >
+1002
+</property>
+<property type="string" id="rule" mode="w" >
+(getattr(evt, 'eventClassKey', '') or '').startswith('MW|WrongCredentials')
```
cluther: This rule can be removed once the mapping's eventClassKey is changed to _MW|WrongCredentials_.

PR #646, opened 2016-11-04
Only load the assembly version needed for the instance

dbouchillon:
Fixes ZEN-25851

Multiple versions of sql server can co-exist.  we should only load
the assembly for the version of the instance we're modeling/monitoring.
This will also display the current version

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/utils.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/utils.py
@@ -136,8 +136,30 @@ def filter_sql_stdout(val):
     return filter(lambda x: x!="LogonUser succedded", val)
 
 
-def getSQLAssembly():
+def getSQLAssembly(version=None):
+    """Return only the powershell assembly version needed for the instance.
+    These statements are prepended to any sql server connections.
+    SQL server 2012 and 2014 can use the same version.
+    If no version is given then return all and hope the first one that
+    loads is correct.
+    Multiple versions of sql server can co-exist on the same machine.
 
+    example using sql server 11 or 12
+    try {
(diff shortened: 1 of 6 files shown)


dbouchillon:
bump

ZenPacks/zenoss/Microsoft/Windows/utils.py
```diff
@@ -176,32 +185,49 @@ def getSQLAssembly():
...
+        sqlConnection.append("try{")
+        sqlConnection.append(MSSQL_CONNECTION_INFO.get(version))
+        sqlConnection.append("}catch{")
+        sqlConnection.append(ASSEMBLY_LOAD_ERROR)
+        sqlConnection.append("}")
+        sqlConnection.append("try{")
+        sqlConnection.append(MSSQL_SMO.get(version))
+        sqlConnection.append("}catch{")
+        sqlConnection.append(ASSEMBLY_LOAD_ERROR)
+        sqlConnection.append("}")
```
pcarinhas: Can these be string formatted so that the code structure is more apparent?
Otherwise comments would be helpful in understanding this.

ZenPacks/zenoss/Microsoft/Windows/utils.py
```diff
@@ -136,8 +136,12 @@ def filter_sql_stdout(val):
     return filter(lambda x: x!="LogonUser succedded", val)
 
 
-def getSQLAssembly():
+def getSQLAssembly(version=None):
+    """Return only the assembly version needed for the instance
```
pcarinhas: Fix the formatting and punctuation here, looks like a run-on.

PR #663, opened 2017-01-11
Initial conversion to ZenPackLib 2.0

j053ph4:
- ZenPacks.zenoss.ZenPackLib is now a dependency and must be installed prior to installation of this ZP
-RRD Templates still reside in objects.xml pending YAML conversion
validation


ZenPacks/zenoss/Microsoft/Windows/__init__.py
```diff
@@ -37,12 +37,11 @@
 OLD_DEVTYPE_PROTOCOL = 'WMI'
 
 
-from . import zenpacklib
+from ZenPacks.zenoss.ZenPackLib import zenpacklib
 
-# CFG is necessary when using zenpacklib.TestCase.
-CFG = zenpacklib.load_yaml()
+CFG = zenpacklib.load_yaml(verbose=True)
```
dbouchillon: let's go ahead and add in the path and filename so it loads faster

ZenPacks/zenoss/Microsoft/Windows/tests/test_AddOSRelations.py
```diff
@@ -101,3 +104,17 @@ def test_unneeded(self):
...
+    from unittest import TestSuite, makeSuite
+    suite = TestSuite()
+    suite.addTest(makeSuite(migrateTests))
+    return suite
+
+
+if __name__ == "__main__":
+    from zope.testrunner.runner import Runner
+    runner = Runner(found_suites=[test_suite()])
+    runner.run()
```
dbouchillon: can you add a newline?

PR #666, opened 2017-01-12
Convert Event Classes and Mappings to YAML (Please merge first)

j053ph4:
- Moved all event classes and mappings from objects.xml to YAML


cluther:
FYI: I wouldn't both with a pull request or code review to merge into a feature branch.

PR #684, opened 2017-01-24
Do not run service indexing job on temporary templates

dbouchillon:
Fixes ZPS-570

* Check for -new, -backup, -preupgrade so that we do not start job


ZenPacks/zenoss/Microsoft/Windows/handlers.py
```diff
@@ -1,25 +1,33 @@
...
 DEFAULT_SERVICE = '/zport/dmd/Devices/Server/Microsoft/rrdTemplates/WinService/datasources/DefaultService'
 
 
 @adapter(ServiceDataSource, IObjectWillBeMovedEvent)
 def onServiceDataSourceMoved(ob, event):
     if isinstance(event, ObjectWillBeRemovedEvent):
         dmd = ob.getDmdRoot("Devices")
-        if dmd:
+        try:
+            template = ob.getPrimaryUrlPath().split('/')[-3]
```
joshw: could this be done as ob.rrdTemplate().id, or something like that?  Might be a little bit clearer?

PR #687, opened 2017-01-26
MSSQL Database state should reflect instance status (Fixes ZPS-299)

j053ph4:
- Fixes ZPS-299
- Updated WinSQLInstance getStatus method to return 1 if any events
- Updated WinSQLDatabase "getState" method to first check WinSQLInstance
state
- Updated ShellDataSource to generate events for every affected
component when a query fails
- added built-in "status" datasource to WinDatabase template so that
"cacheRRDValue" will work


j053ph4:
bump please

PR #690, opened 2017-01-30
Add event for each failed datasource (ZPS-566)

j053ph4:
- Fixes ZPS-566
- Update EventLogDataSource to generate an event for each failed data
source, so that failures are tracked on a per-datasource basis

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/datasources/EventLogDataSource.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/datasources/EventLogDataSource.py
@@ -152,14 +152,14 @@ class EventLogInfo(InfoBase):
                 time_match = re.match('(.*TimeCreated)(\[.*?\])(.*)', filter_text)
                 if time_match:
                     # easy to replace with our time filter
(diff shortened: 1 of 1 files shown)


ZenPacks/zenoss/Microsoft/Windows/datasources/EventLogDataSource.py
```diff
@@ -334,13 +334,14 @@ def onSuccess(self, results, config):
...
             data['events'].append(self._makeEvent(evt, config))
 
-        data['events'].append({
-            'device': config.id,
-            'summary': 'Windows EventLog: successful event collection',
-            'severity': ZenEventClasses.Clear,
-            'eventKey': 'WindowsEventCollection',
-            'eventClassKey': 'WindowsEventLogSuccess',
-        })
+        for ds in config.datasources[0]:
```
ssOleg: is it ok "for ds in config.datasources[0]" ?

j053ph4: good catch thank you

PR #693, opened 2017-01-31
Ensure that status will display up/down

dbouchillon:
Fixes ZPS-632

* Underlying class depends upon monitoredStartModes being populated.
  Populate it with the appropriate start modes depending upon how
  service is being monitored
* Also make it clear that the checkbox for usermonitor does not enable/disable monitoring
* Make sure that DefaultService has startmodes attribute


dbouchillon:
bump

PR #694, opened 2017-01-31
Make sure datasource has startmode attribute

dbouchillon:
Fixes ZPS-661

* Some templates may end up with a bad datasource that is not a
  Windows Service datasource.  Skip them


dbouchillon:
bump please.  need to get hotfix out

PR #696, opened 2017-02-01
Fix Processes GUI display (ZPS-572)

j053ph4:
- Fixes ZPS-572
- Reverted earlier partial fix (platform patch)
- Updated setup.py dependency to ZenPackLib 2.0.2 (forthcoming), which
contains the actual fix
- added ResetOSProcesses migrate script for existing OSProcess objects
- removed "getClassObject" method from OSProcess override (no longer
needed)

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/migrate/ResetOSProcesses.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/migrate/ResetOSProcesses.py
@@ -1,0 +1,75 @@
+##############################################################################
+#
+# Copyright (C) Zenoss, Inc. 2015, all rights reserved.
+#
+# This content is made available according to terms specified in
+# License.zenoss under the directory where your Zenoss product is installed.
+#
+##############################################################################
+
+"""Reset meta_type and portal_type for objects that need it."""
+
+# Logging
+import logging
+LOG = logging.getLogger("zen.Microsoft.Windows.migrate.{}".format(__name__))
+
+# Zenoss Imports
+from zope.event import notify
+from Products.ZenModel.migrate.Migrate import Version
+from Products.ZenModel.ZenPack import ZenPackMigration
+from Products.Zuul.catalog.events import IndexingEvent
+from Products.Zuul.interfaces import ICatalogTool
+
+# ZenPack Imports
+from ZenPacks.zenoss.Microsoft.Windows import progresslog
+
+
+# If the migration takes longer than this interval, a running progress
+# showing elapsed and estimated remaining time will be logged this
+# often. The value is in seconds.
+PROGRESS_LOG_INTERVAL = 10
+
+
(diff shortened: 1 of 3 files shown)


j053ph4:
bump please

ZenPacks/zenoss/Microsoft/Windows/migrate/ResetOSProcesses.py
```diff
@@ -0,0 +1,78 @@
...
+# showing elapsed and estimated remaining time will be logged this
+# often. The value is in seconds.
+PROGRESS_LOG_INTERVAL = 10
+
+
+class ResetOSProcesses(ZenPackMigration):
+    version = Version(2, 7, 0)
+
+    def migrate(self, pack):
+        LOG.info("searching for objects")
```
dbouchillon: This should be more descriptive for the objects for which it's searching.  Or it may actually be unnecessary if there are no Processes to migrate

ZenPacks/zenoss/Microsoft/Windows/migrate/ResetOSProcesses.py
```diff
@@ -0,0 +1,78 @@
...
+        results = ICatalogTool(pack.getDmdRoot("Devices")).search(types=(
+            'ZenPacks.zenoss.Microsoft.Windows.OSProcess.OSProcess',
+            ))
+
+        if not results.total:
+            LOG.info("no objects to migrate")
```
dbouchillon: See above for the "searching for objects".  If there are no objects to migrate then there's no real reason to inform the user.

ZenPacks/zenoss/Microsoft/Windows/migrate/ResetOSProcesses.py
```diff
@@ -0,0 +1,78 @@
...
+            return
+
+        LOG.info("starting: %s total objects", results.total)
```
dbouchillon: I think making this message more descriptive would be a better way to go instead of the two infos above.  Also, you should use `.format` instead.

`"Found {} Processes that may require indexing".format(results.total)`

ZenPacks/zenoss/Microsoft/Windows/migrate/ResetOSProcesses.py
```diff
@@ -0,0 +1,78 @@
...
+        objects_migrated = 0
+
+        for result in results:
+            if self.migrate_result(result):
+                objects_migrated += 1
+
+            progress.increment()
+
+        LOG.info(
+            "finished: %s of %s objects required migration",
```
dbouchillon: Also be more descriptive and use capitalization.  use `.format`

maybe `"Finished indexing {} of {} Processes"`.format(objects_migrated, results.total)`

j053ph4:
OK I updated the those logging messages

PR #700, opened 2017-02-03
Use new txwinrm lib

dbouchillon:
Fixes ZPS-741


dbouchillon:
self-merging, txwinrm was reviewed in https://github.com/zenoss/txwinrm/pull/98

PR #712, opened 2017-02-09
Fix Service Status Display (ZPS-762)

j053ph4:
- Fixes ZPS-762
- Rebase WinService on ZPL proxy class
- Migrated class properties to YAML
- Set dependency on ZenPackLib 2.0.2


dbouchillon:
Do we need to migrate the services during installation?

j053ph4:
I thought it would be (and initially updated a migrate script to do so), but this issue was fixed without migrating --- do you think we should just in case?

j053ph4:
OK updated the migrate script

PR #724, opened 2017-02-14
Fix ShellDataSource status datapoints (ZPS-752)

j053ph4:
- Fixes ZPS-752
- Fixes ZPS-754
- Updated attributes of dummy datapoint config to be platform dependent


j053ph4:
bumping for great justice

PR #737, opened 2017-02-22
Accurately determine device status (ZPS-873)

j053ph4:
- Fixes ZPS-873
- Override Device "getStatus" method to only use events in
"/Status/Ping" or "/Status/Winrm/Ping" to determine device status
indicator.

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/Device.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/Device.py
@@ -23,6 +23,12 @@ class Device(schema.Device):
     Model class for a Windows operating system device.
     '''
 
+    def getStatus(self, statusclass='/Status/Ping', **kwargs):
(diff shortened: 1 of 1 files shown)


joshw:
(approval feedback left- slack integration isn't picking it up)

j053ph4:
The "getPingStatus" method does call "getStatus" with various arguments, and the reason for overriding getPingStatus was so that it will include the '/Status/Winrm/Ping" method in the list of eventclasses it would otherwise use (expanding its capability).  

Overriding getStatus, though, corrects the opposite problem (device down being indicated by anything under /Status)  by restricting the allowed event classes to /Status/Ping and /Status/Winrm/Ping.

The broader issue is that many of the various Windows event classes live under /Status/Winrm, such as auth failures and the like, and it's undesirable for these failures to indicate that the device is down.

PR #741, opened 2017-02-23
Default zDBInstances user and password should be None by default

j053ph4:
- Changed user/pass to None instead of "" (empty string), so that
default credentials are used


dbouchillon:
is this for a specific jira?

j053ph4:
no (but not sure about that) -- i just noticed it like the event transform issue

dbouchillon:
ok.  then let's leave it like it is.  if something crops up, then we can look at it again

PR #756, opened 2017-03-01
Set wide shell width in case of long db names, cluster also

dbouchillon:
Fixes ZPS-944

Set wide shell width so that lines do not wrap around

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/datasources/ShellDataSource.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/datasources/ShellDataSource.py
@@ -48,13 +48,13 @@ from ZenPacks.zenoss.Microsoft.Windows.utils import filter_sql_stdout, \
     parseDBUserNamePass, getSQLAssembly
 from ..utils import (
     check_for_network_error, pipejoin, sizeof_fmt, cluster_state_value,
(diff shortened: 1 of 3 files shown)


dbouchillon:
bump

dbouchillon:
bump

ZenPacks/zenoss/Microsoft/Windows/datasources/ShellDataSource.py
```diff
@@ -517,7 +522,10 @@ def build_command_line(self, sqlConnection):
         jobs_sqlConnection.append("'|CurrentRunStatus:'$job.CurrentRunStatus;")
         jobs_sqlConnection.append("}}")
         script = "\"& {{{}}}\"".format(
-            ''.join(getSQLAssembly(sqlConnection.version) + sqlConnection.sqlConnection + jobs_sqlConnection))
+            ''.join([BUFFER_SIZE] +
+                    getSQLAssembly(sqlConnection.version) +
+                    sqlConnection.sqlConnection +
+                    jobs_sqlConnection))
```
yichi-lu: got this while modeling: ERROR zen.PythonClient: Error in zenoss.winrm.WinMSSQL: cannot concatenate 'str' and 'list' objects.

dbouchillon:
Weird i didn't get that when i was initially testing.  getting it now though.  updated code

PR #760, opened 2017-03-01
Add link to storage disk/lun for hard disk

yichi-lu:
Fixes ZPS-939

If the link to the storage disk/lun can be found, add the link
to hard disk details.

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/zenpack.yaml
+++ b/ZenPacks/zenoss/Microsoft/Windows/zenpack.yaml
@@ -448,6 +448,15 @@ classes:
         type: int
       instance_name:
         details_display: false
+      storage_disk_lun:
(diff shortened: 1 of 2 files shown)


ZenPacks/zenoss/Microsoft/Windows/zenpack.yaml
```diff
@@ -448,6 +448,15 @@ classes:
         type: int
       instance_name:
         details_display: false
+      storage_disk_lun:
+        # link to storage disk/virtual drive
+        # if the hard disk is actually mapped to it
+        label: Storage Disk/LUN
+        type: entity
+        grid_display: false
```
dbouchillon: grid_display is false by default for HardDisk properties.  no need to add it here

PR #764, opened 2017-03-03
Removed references to localization support

dbouchillon:
Fixes ZPS-945

We are not going to advertise localization support


dbouchillon:
bump

PR #765, opened 2017-03-06
Check both ip and clustername to make sure device doesn't exist

dbouchillon:
Fixes ZPS-917


joshw:
looks good to me.  only suggestion i would have would be to switch the order, and save the dns lookup, if you considered the dnsname to be an adequately unique ID.  Since the IP is coming from DNS anyway, it's possible that you might be able to do that.

dbouchillon:
ok, switched search order

PR #766, opened 2017-03-06
Add Paused as a valid WinService status

dbouchillon:
Fixes ZPS-942

* Allow for Paused state of service.  Service may be ok if it's just paused with no need to alert

update readme


j053ph4:
would you need to patch this into core WinService.py as well? (for zMonitoredStartModes)?

dbouchillon:
nope,  start modes are different.  start mode is 'Auto', 'Manual', 'Disabled'.  current state would be running, stopped, paused.  WinService does not give options for state

PR #773, opened 2017-03-09
Add more information to events to track down issue

dbouchillon:
Fixes ZPS-831

also add more logging info for dcdiag and custom command debugging

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/datasources/PerfmonDataSource.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/datasources/PerfmonDataSource.py
@@ -41,6 +41,7 @@ from ZenPacks.zenoss.PythonCollector.datasources.PythonDataSource import (
 
 from ..twisted_utils import add_timeout
 from ..txwinrm_utils import ConnectionInfoProperties, createConnectionInfo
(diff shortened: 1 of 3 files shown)


dbouchillon:
bump

ZenPacks/zenoss/Microsoft/Windows/datasources/PerfmonDataSource.py
```diff
@@ -798,37 +798,49 @@ def cleanup(self, config):
...
+        plugin_classname = datasources[0].plugin_classname
         if 'Password expired' in errorMessage:
             PERSISTER.add_event(self.config.id, {
                 'device': self.config.id,
                 'severity': ZenEventClasses.Critical,
                 'eventClassKey': 'MW|PasswordExpired',
                 'summary': errorMessage,
-                'ipAddress': self.config.manageIp})
+                'ipAddress': self.config.manageIp,
+                'plugin_classname': plugin_classname,
```
cluther: Rather than writing out this same change all over the place, what about making a utility method on a base plugin class or something like that to automatically add these generally-useful fields?

dbouchillon:
ok, changes made

PR #777, opened 2017-03-15
Add Cluster nodes automatically (ZPS-577)

j053ph4:
Fixes ZPS-577

Updated "create_device" method to use facade instead of dmd transaction
(submits as job).  Also modified to set zProperties per the cluster
settings.


dbouchillon:
approved for merging

PR #780, opened 2017-03-16
update with new txwinrm

dbouchillon:
Fixes ZPS-877


dbouchillon:
merged as the only change was new txwinrm lib, approved in https://github.com/zenoss/txwinrm/pull/114

PR #782, opened 2017-03-16
Revert to legacy rrdPath (ZPS-292)

j053ph4:
- Fixes ZPS-292
- changes rrdPath method on some component classes to revert to older
behavior for Zenoss 4.x
- Since issue affects multiple classes besides interface, these were
updated as well
- Although this update does not recover lost data, it will prevent
further loss while eliminating it altogether for direct upgrades from
2.6.2


dbouchillon:
Approved

PR #785, opened 2017-03-20
Suppress uncatalog errors during upgrade (ZPS-984)

j053ph4:
- Fixes ZPS-984
- These errors occur as a downstream effect of "winrmservices" component
removal via relation deletion during migration.  Since deleting the
relation causes the related components to deleted simultaneously, this
causes a spurious uncatalogObject warning that can be disregarded.  The
"suppress_events" flag has been added, which should prevent these
warnings from being seen by the user.


dbouchillon:
approved

PR #786, opened 2017-03-20
Logging should include device and component where relevant (ZPS-573)

j053ph4:
- Fixes ZPS-573
- Updated log messaging for custom datasources to include the
device/component ids where available.


dbouchillon:
approved

PR #788, opened 2017-03-21
ZPS-1149: Fix tracebacks in zenpython.

mihnatenko:
Fixes [ZPS-1149](https://jira.zenoss.com/browse/ZPS-1149).

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/datasources/PerfmonDataSource.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/datasources/PerfmonDataSource.py
@@ -1,6 +1,6 @@
 ##############################################################################
 #
-# Copyright (C) Zenoss, Inc. 2013-2016, all rights reserved.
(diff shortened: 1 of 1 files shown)


ZenPacks/zenoss/Microsoft/Windows/datasources/PerfmonDataSource.py
```diff
@@ -197,8 +197,17 @@ def __init__(self, dsconf, num_commands):
...
         """Create initial set of commands according to the number supplied."""
         self.num_commands = num_commands
-        return [LongCommandClient(createConnectionInfo(self.dsconf))
-                for i in xrange(num_commands)]
+
+        def _long_cmd_client_wrap():
+            try:
+                conn_info = createConnectionInfo(self.dsconf)
+                return LongCommandClient(conn_info)
+            except Exception as e:
```
ssOleg: Is there a way to catch specific exceptions, and probably LOG correct message for all cases?

PR #794, opened 2017-03-22
ZPS-574: Report Systems, Groups, Locations.

byshovets:
Fixes https://jira.zenoss.com/browse/ZPS-574, https://jira.zenoss.com/browse/ZPS-953

* Schema fixes for Device and OS Process
* analytics-bundle.zip recreated.


dbouchillon:
i get this error when i try to run the bundle creator on a device that is in a location, system, and a group:

```
$: ./create-analytics-bundle --folder="Microsoft Windows ZenPack" --domain="Microsoft Windows Domain"  --device=windows1
Traceback (most recent call last):
  File "./create-analytics-bundle", line 928, in <module>
    creator.run()
  File "./create-analytics-bundle", line 69, in run
    self.create_zenpack_domain()
  File "./create-analytics-bundle", line 193, in create_zenpack_domain
    update_schema(schema, self.device)
  File "./create-analytics-bundle", line 564, in update_schema
    un_camel(relName))
  File "/x/ZenPacks.zenoss.ZenETL/ZenPacks/zenoss/ZenETL/reportable.py", line 1293, in createReferenceProperty
    propValue = IReportable(entity).sid if entity else None
  File "/opt/zenoss/lib/python/zope/component/hookable.py", line 33, in __call__
    return self.__implementation(*args, **kw)
  File "/opt/zenoss/lib/python/zope/component/hooks.py", line 115, in adapter_hook
    return siteinfo.adapter_hook(interface, object, name, default)
TypeError: __init__() takes exactly 3 arguments (2 given)
```
i think we need to skip over the location relation when we're looping through the relations.  around line 551

byshovets:
@dbouchillon thank you. I updated create-analytics-bundle script, so it skips "location" property. Systems and Groups would not be a problem because they are "TooMany" relations.

PR #797, opened 2017-03-22
Don't blow up even if modeling a non-supported device

dbouchillon:
Fixes ZPS-1205

win2003 is not supported, but we should not traceback if modeling one


dbouchillon:
bump.  another simple one

PR #800, opened 2017-03-24
Remove user visibility of fix for ZPS-1204

j053ph4:
- Updated fix for ZPS-1204
- Changes previously introduced zProperty into an attribute stored on
dmd, set by install migration script that determines its value.
- Removes user visibility while still preventing the data loss
associated with rrdPath changes from zenpacklib conversion

### DIFF
--- a/README.mediawiki
+++ b/README.mediawiki
@@ -794,11 +794,6 @@ The [[#Adding a Windows Device]] steps shown above are for the simplest case of 
 
 {{note}} HyperV and MicrosoftWindows ZenPacks share krb5.conf file as well as tools for sending/receiving data. Therefore if either HyperV or Windows device has a correct zWinKDC setting, it will be used for another device as well.
 
-;zWinUseLegacyRRDPath
-: Whether to use older directory structure to manage RRD files.  Set this to True if upgrading directly from 2.5.x of this ZenPack if using Zenoss versions earlier than 5.x.  This setting has no effect on Zenoss versions 5.x and above.
-
-{{note}} The migration to ZenPackLib in version 2.6 of this ZenPack introduced an issue leading to apparent data loss in RRD graphs for filesystems, interfaces, and other components.  Since ZenPackLib uses a different directory heirarchy for RRD storage, component graphs after upgrade became empty (since the file paths changed).  This setting mitigates that issue for users upgrading directly from 2.5.x versions of this ZenPack.  The attempt is made during ZenPack installation to choose the correct version of this value.
-
 <br clear=all>
 
 === Configuring MSSQL Server Modeling/Monitoring ===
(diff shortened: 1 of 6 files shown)


ZenPacks/zenoss/Microsoft/Windows/migrate/MigrateRRDPaths.py
```diff
@@ -28,10 +28,12 @@ class MigrateRRDPaths(ZenPackMigration):
...
+        # setting the default
+        pack.dmd.windows_using_legacy_rrd_paths = False
         if pack.prevZenPackVersion is None:
             return
         installed_version = pkg_resources.parse_version(pack.prevZenPackVersion)
         # we only want to set this if upgrading directly from 2.5.x
         if installed_version < pkg_resources.parse_version('2.6.0'):
-            LOG.info('Setting zWinUseLegacyRRDPath to True we are upgrading directly from 2.5.x')
-            pack.dmd.Devices.setZenProperty('zWinUseLegacyRRDPath', True)
+            LOG.info('Using legacy RRDPath since we are upgrading directly from 2.5.x')
```
cluther: This log might not be true. They could be upgrading from something earlier than 2.5.x too.

What is the user supposed to do with this information? If there's no reason for the user to know about this, or anything for the user to do about it, then there may be no need to log it at all.

README.mediawiki
```diff
@@ -656,6 +656,8 @@ yum -y install krb5-workstation
 
 == Usage ==
 
+{{note}} The migration to ZenPackLib in version 2.6 of this ZenPack introduced an issue leading to apparent data loss in RRD graphs for filesystems, interfaces, and other components.  Since ZenPackLib uses a different directory heirarchy for RRD storage, component graphs after upgrade became empty (since the file paths changed).  Beginning in 2.7.0, an attribute called "windows_using_legacy_rrd_paths" is set to True during ZenPack installation if the previously installed version is earlier than 2.6.  This variable may need to be manually set (via zendmd) in rare scenarios such as rebuilding a Zenoss 4.x system from scratch (after updating from 2.5 to 2.7 of this ZenPack).  Also note that this issue does not affect Zenoss 5 and above.
```
cluther: Putting this note as the very top thing in the _Usage_ section seem to be placing too high an importance on it. It'd probably better belong in a troubleshooting section such as, "Why is all of my data gone after I upgrade?" to help people who upgraded to 2.6 and want to know what happened.

I'd also highlight very clearly right up front that this information is only applicable to those running Zenoss 4.

cluther: I don't think it's a good idea to even document dmd.windows_using_legacy_rrd_paths. I don't think there are even rare scenarios in which someone would change its value.

PR #801, opened 2017-03-24
Ensure zDBInstances is stored as a string (ZPS-870)

j053ph4:
Just a note that I was unable to actually duplicate the issue, but I suspect it is limited to long-running platform installations that have been through many upgrades of the Windows ZP (dating back to at least 2.1.0).  In that sense this fix is speculative, but I don't believe it will cause any harm.

- Fixes ZPS-870
- Updated MigrateDBInstances to test all device classes and devices for
invalid values
- Issue may result from missed migration targets in 2.1 version (so a
long-standing but mostly dormant issue)


dbouchillon:
approved

PR #816, opened 2017-03-29
Revise and simplify auth-related event handling (ZPS-562)

j053ph4:
- Fixes ZPS-562
- Ensure that authentication-related console events are cleared
- Revised structure of event classes and mappings to more simply handle
various use cases
- Prevent and ensure removal of duplicate clear events (which was the
root of the problem)
- Updated testAuthErrEvents unit test


dbouchillon:
approved

PR #823, opened 2017-03-31
Prevent AttributeError during migrate from terminating installation

j053ph4:
- Updated RemoveWinRMServices to continue if winrmservices relation is
not found

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/migrate/RemoveWinRMServices.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/migrate/RemoveWinRMServices.py
@@ -21,20 +21,28 @@ import logging
 # Zenoss Imports
 from Products.ZenModel.migrate.Migrate import Version
 from Products.ZenModel.ZenPack import ZenPackMigration
+from Products.Zuul.interfaces import ICatalogTool
 
 LOG = logging.getLogger('zen.MicrosoftWindows')
 
 
 class RemoveWinRMServices(ZenPackMigration):
-    version = Version(2, 7, 0)
+    version = Version(2, 7, 1)
 
     def migrate(self, pack):
-        org = pack.dmd.Devices.getOrganizer('/Server/Microsoft')
-        devices = org.getSubDevices()
-        device_count = len(devices)
-        if device_count:
-            LOG.info('Attempting to remove incompatible Windows Services from {} device{}.'
-                     .format(device_count, 's' if device_count > 1 else ''))
-            for device in devices:
+
+        results = ICatalogTool(pack.getDmdRoot("Devices")).search(types=(
+            'ZenPacks.zenoss.Microsoft.Windows.BaseDevice.BaseDevice',
+        ))
+
+        if not results.total:
+            return
+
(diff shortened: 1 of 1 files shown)


ZenPacks/zenoss/Microsoft/Windows/migrate/RemoveWinRMServices.py
```diff
@@ -21,20 +21,30 @@
...
-                device.os.removeRelation('winrmservices', suppress_events=True)
+
+        results = ICatalogTool(pack.getDmdRoot("Devices")).search(types=(
+            'ZenPacks.zenoss.Microsoft.Windows.BaseDevice.BaseDevice',
+        ))
+
+        if not results.total:
+            return
+
+        LOG.info('Attempting to remove incompatible Windows Services from {} device{}.'
```
cluther: "Attempting to remove" seems superfluous. How about just "Removing incompatible ...".

As a general best practice I would recommend sticking with old-style string formatting as supported by logging instead of string.format for log messages. It has the advantage of only doing the string formatting when the log message would actually be emitted based on logging configuration. That makes no significant difference here, but it can make significant difference in other places, and this way we keep it consistent.

ZenPacks/zenoss/Microsoft/Windows/migrate/RemoveWinRMServices.py
```diff
@@ -21,20 +21,30 @@
...
-        devices = org.getSubDevices()
-        device_count = len(devices)
-        if device_count:
-            LOG.info('Attempting to remove incompatible Windows Services from {} device{}.'
-                     .format(device_count, 's' if device_count > 1 else ''))
-            for device in devices:
-                device.os.removeRelation('winrmservices', suppress_events=True)
+
+        results = ICatalogTool(pack.getDmdRoot("Devices")).search(types=(
+            'ZenPacks.zenoss.Microsoft.Windows.BaseDevice.BaseDevice',
```
cluther: Have you verified that this does return the expected brains? Sometimes searching the global catalog by type has unexpected results when searching for parent classes. This is because only so many parents up the hierarchy are actually indexed for performance reasons. With multiple inheritance you don't quite know which branches will be followed, and therefore which parent classes will actually be indexed.

So you need to be sure to perform this same search on a system with some Windows devices to make sure they do get returned.

ZenPacks/zenoss/Microsoft/Windows/migrate/RemoveWinRMServices.py
```diff
@@ -21,20 +21,30 @@
...
+        if not results.total:
+            return
+
+        LOG.info('Attempting to remove incompatible Windows Services from {} device{}.'
+                     .format(results.total, 's' if results.total > 1 else ''))
+
+        for r in results:
+            try:
+                device = r.getObject()
                 device.os.buildRelations()
```
cluther: This buildRelations() call is not necessary. It would be to handle the case of the winrmservices relationship not yet existing on device.os, but that case is already handled by the device.os.removeRelation call already being inside of an exception handler that ignores all exceptions.

PR #826, opened 2017-04-10
Remove 'datasources' field from event details (ZPS-1293)

j053ph4:
- Fixes ZPS-1293
- Remove 'datasources' field from generic events, since it occupies too
much screen real estate in the event details GUI


dbouchillon:
approved

PR #827, opened 2017-04-10
Serial number could be None, which has no strip() method

dbouchillon:
Fixes ZPS-1279

Also added unit test

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/modeler/plugins/zenoss/winrm/HardDisks.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/modeler/plugins/zenoss/winrm/HardDisks.py
@@ -75,9 +75,17 @@ class HardDisks(WinRMPlugin):
                 size = int(drive.Size)
             except TypeError:
                 size = 0
-            serialNumber = drive.SerialNumber.strip() if hasattr(drive, 'SerialNumber') else ''
+
+            # drive.SerialNumber could be None.  let's make it ''
+            serialNumber = ''
+            if hasattr(drive, 'SerialNumber'):
+                if drive.SerialNumber is None:
+                    drive.SerialNumber = ''
+                serialNumber = drive.SerialNumber.strip()
             product_key = MultiArgs(drive.Model, drive.Manufacturer)
-            capabilities = drive.CapabilityDescriptions if hasattr(drive, 'CapabilityDescriptions') else ''
+            capabilities = ''
+            if hasattr(drive, 'CapabilityDescriptions') and drive.CapabilityDescriptions:
+                capabilities = drive.CapabilityDescriptions
             rm.append(self.objectMap({
                 'id': self.prepId(drive.PNPDeviceID),
                 'title': drive.Caption,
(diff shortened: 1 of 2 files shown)


dbouchillon:
bump

ZenPacks/zenoss/Microsoft/Windows/modeler/plugins/zenoss/winrm/HardDisks.py
```diff
@@ -75,9 +75,17 @@ def process(self, device, results, log):
                 size = int(drive.Size)
             except TypeError:
                 size = 0
-            serialNumber = drive.SerialNumber.strip() if hasattr(drive, 'SerialNumber') else ''
+
+            if drive.SerialNumber is None:
```
itshane: It will cause an AttributeError when the SerialNumber is not in the drive.

itshane: I would suggest something like this:
```
            serialNumber = ''
            if hasattr(drive, 'SerialNumber') and drive.SerialNumber:
                serialNumber = drive.SerialNumber.strip()
```

dbouchillon:
ok, made changes requested

SerialNumber can come in and be evaled as None, so we can just set it to ''

ZenPacks/zenoss/Microsoft/Windows/modeler/plugins/zenoss/winrm/HardDisks.py
```diff
@@ -128,7 +136,9 @@ def make_disk_ids(disk_drive):
     disk_ids.append(pnpDevice)
 
     # Append SerialNumber
-    serial = disk_drive.SerialNumber.strip() if hasattr(disk_drive, 'SerialNumber') else None
+    serial = None
+    if hasattr(disk_drive, 'SerialNumber'):
+        serial = disk_drive.SerialNumber.strip()
```
itshane: What if disk_drive.SerialNumber is None?

ZenPacks/zenoss/Microsoft/Windows/modeler/plugins/zenoss/winrm/HardDisks.py
```diff
@@ -75,9 +75,17 @@ def process(self, device, results, log):
                 size = int(drive.Size)
             except TypeError:
                 size = 0
-            serialNumber = drive.SerialNumber.strip() if hasattr(drive, 'SerialNumber') else ''
+
+            # drive.SerialNumber could be None.  let's make it ''
+            serialNumber = ''
+            if hasattr(drive, 'SerialNumber'):
+                if drive.SerialNumber is None:
```
itshane: You could reduce to two lines:
```
if drive.SerialNumber is not None:
    serialNumber = drive.SerialNumber.strip()
```

ZenPacks/zenoss/Microsoft/Windows/modeler/plugins/zenoss/winrm/HardDisks.py
dbouchillon: it can't be because if it is set to None initially, I change it to '' before it comes into this function

ZenPacks/zenoss/Microsoft/Windows/modeler/plugins/zenoss/winrm/HardDisks.py
dbouchillon: i change drive.SerialNumber to '' if it comes in as None so that it will work correctly in make_disk_ids

dbouchillon: Also, the platform default for serialNumber is ''

itshane:
Please squash the commits

PR #832, opened 2017-04-13
Inherit WinService severity from ServiceClass by default

j053ph4:
- WinService should attempt to set failSeverity by acquisition first
before using the default.


dbouchillon:
approved

PR #837, opened 2017-04-17
Explicit timestamp for WinService state datapoint (ZPS-1341)

j053ph4:
- Fixes ZPS-1341
- Update ServiceDataSource to provide timestamp along with status
datapoint.


dbouchillon:
looks good

PR #838, opened 2017-04-17
Make getAffectedServices smarter

dbouchillon:
Fixes ZPS-1342

This should reduce the overall number of services that need to be
reindexed.  We'll look at the template id to see if it is an actual
Service Name.  Next we'll make a query out of the in_exclusions regex
to send into ICatalogTool.  There's no need to index services that
would not be affected.


dbouchillon:
bump

PR #839, opened 2017-04-25
write out database name in case no counter data returned

dbouchillon:
Fixes ZPS-1271

update readme


dbouchillon:
bump.  this is very simple one

PR #841, opened 2017-05-01
remove winrmservices relation from OperatingSystem

dbouchillon:
fixes ZPS-1442

* winrmservices should no longer be a relationship on a windows device
* remove it from the os and from the modeler plugin.
* also no need to try and remove in migrate script

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/jobs.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/jobs.py
@@ -12,7 +12,6 @@ Zenpack's jobs.
 """
 
 from ZODB.transact import transact
-from transaction import commit
 from zope.event import notify
 from Products.Jobber.jobs import Job
 from Products.Zuul.catalog.events import IndexingEvent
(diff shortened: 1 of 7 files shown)


ZenPacks/zenoss/Microsoft/Windows/zenpack.yaml
```diff
@@ -47,7 +47,6 @@ zProperties:
     default: false
 
 class_relationships:
-  - ZenPacks.zenoss.Microsoft.Windows.OperatingSystem.OperatingSystem(winrmservices) 1:MC (os)WinService
```
cluther: I don't think you can do this. You'll certainly end up in a situation where the relationship schema doesn't match the relationship reality. It's not so bad when the schema contains relationships that an object doesn't, but it's very bad when the object contains a relationship that doesn't exist in the schema.

You need to leave vestigial relationships around forever. You can empty them, but they need to remain.

dbouchillon:
changes made so that we keep winrmservices relationship and remove any old winrmservice correctly

dbouchillon:
ok, changes made per our discussion

ZenPacks/zenoss/Microsoft/Windows/jobs.py
```diff
@@ -75,17 +76,19 @@ def getJobDescription(self, *args, **kwargs):
...
-        for device in org.getSubDevicesGen():
-            finished = False
-            devices_cleaned += 1
-            # This could take several passes
-            while not finished:
-                self.remove_winrmservices(device)
-                finished = False if device.componentSearch(meta_type='WinRMService') else True
-            if not devices_cleaned % 25:
-                self.log.info('Committing devices with unclean winrmservices removed.')
+        results = ICatalogTool(org).search(WinService, query=Eq('meta_type', 'WinRMService'))
```
cluther: I don't think there's much performance benefit to using the catalog here instead of just iterating over all Windows devices and their os.winrmservices relationship. I think there's a chance that there will be objects in the os.winrmservices relationship that have a meta_type that isn't WinRMService that will be missed by this approach.

I think an approach like the following would be more likely to get rid of all device.os.winrmservices objects.

```
from Products.Zuul.interfaces import ICatalogTool
from ZenPacks.zenoss.Microsoft.Windows.BaseDevice import BaseDevice

device_results = ICatalogTool(dmd.Device).search(BaseDevice)
for device_result in device_results:
    # remove all device_result.getObject().os.winrmservices objects
```

Also, the `progresslog` module is already in the Windows ZenPack. So you might want to consider using it for this instead of your own modulo progress logging. See the ResetOSProcesses migrate script for an example.

PR #842, opened 2017-05-02
Make sure that $s use the powershell escape(`) if in double quotes

dbouchillon:
Fixes ZPS-1404

we surround a counter with double quotes.  powershell attempts to
evaluate the $ in some counters the same as bash.  use the powershell
escape char (`) so that we can poll these counters


dbouchillon:
bump.  this is a simple one

PR #844, opened 2017-05-09
Improve speed of migrate script

dbouchillon:
Fixes ZPS-1473

Using catalog to find templates and only look at those with a targetPythonClass
of ZenPacks.zenoss.Microsoft.Windows.WinService.  Much faster than looking at
individual services


dbouchillon:
bump

PR #848, opened 2017-05-31
Change is_iis to a boolean, apply during modeling iis components

dbouchillon:
Fixes ZPS-1576

move is_iis to a boolean device property.  set this when IIS
modeler plugin is used


dbouchillon:
bump.  easy one, no windows knowledge necessary

PR #851, opened 2017-06-02
Fix redundant service state metric publishing

cluther:
Add the state datapoint in the plugin's constructor when the task is
built instead of on each collection cycle.

Fixes ZPS-1604.


dbouchillon:
looks good

PR #852, opened 2017-06-07
Handle invalid Size on HardDisk

dbouchillon:
Fixes ZPS-1424

also update unit test


dbouchillon:
bump.  very simple one.

PR #853, opened 2017-06-07
Do not start persister when module is loaded

dbouchillon:
Fixes ZPS-1600

start maintenance when we start collection


dbouchillon:
bump.  another very easy one

PR #857, opened 2017-06-19
send nagios values, events even with error exit code

dbouchillon:
Fixes ZPS-1366

Nagios can have status code of an Error and will still have valid
datapoints.  Honor the dps, and event class/key in this case. update
severity according to nagios status.  update unit tests


dbouchillon:
bump

dbouchillon:
bump please

PR #858, opened 2017-06-22
Make expected receive failure logs debug instead of warning

dbouchillon:
Fixes ZPS-1693

If we know the failures are expected, there's no need for a warning


dbouchillon:
bump.  super simple

PR #859, opened 2017-06-27
Move cluster monitoring to new datasource. Consolidate into one call

dbouchillon:
Fixes ZPS-1697

Cluster monitoring was opening a powershell per component. This moves
that into one single powershell command to return state of all components
that we will parse.  Added unit test to verify onSuccess.

Add cluster datasource


ZenPacks/zenoss/Microsoft/Windows/datasources/ClusterDataSource.py
```diff
@@ -0,0 +1,296 @@
...
+            # ignore any empty lines
+            if len(result) <= 0:
+                continue
+            try:
+                comp, state = result.split('|')
+            except Exception:
+                log.debug('Unable to parse cluster result {} on {}'.format(result, config.id))
+                continue
+            comp = prepId(comp)
+            data['values'][comp]['state'] = cluster_state_value(state), int(time.mktime(time.localtime()))
```
joshw: I think you can just pass 'N' and avoid any issues with time zones on that mktime/localtime thing (i'm not sure, but that might assume that the local timezone is GMT to give you the right value)

dbouchillon: yeah, i copied the time thing from another place.  changing

PR #863, opened 2017-07-11
Reduce load on event server

dbouchillon:
Fixes ZPS-1752

Consolidate kerberos errors into one event.
Add Clear severity to clear events so event server does not need to process them
Update unit tests
IISSiteDataSource should send a clear for sitecheck Running and use the ds severity for not running or unknown
Remove superfluous KerberosAuthenticaion event instances

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ dist
 build
 eggs
 txwinrm
+*.bak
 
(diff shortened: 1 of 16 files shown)


dbouchillon:
Bump.  need to get this out to customer

PR #865, opened 2017-07-14
Update how we show IIS status

dbouchillon:
Fixes ZPS-490

Use datapoint for IIS Site status
remove unnecessary event class mappings


dbouchillon:
bump

PR #867, opened 2017-07-21
Fix 'kerberos' referenced before assignment

cluther:
The fix for this error is zenoss/txwinrm@fd449ea which was released in
txwinrm-1.2.22.

Fixes ZPS-1795.


joshw:
approved

PR #876, opened 2017-08-11
invalidate service object to free up memory

dbouchillon:
Fixes ZPS-1783

also added some updates to the readmes


dbouchillon:
bump

dbouchillon:
bump

PR #877, opened 2017-08-14
Find cluster node correctly

dbouchillon:
Fixes ZPS-1852

We should not match on empty ipaddresses and we should only look
for node within /Server/Microsoft


dbouchillon:
bump

joshw:
approved

PR #886, opened 2017-08-24
update createdevice to create multiple devices with multiple services

dbouchillon:
used to test ZPS-1783


dbouchillon:
This is just to help with testing out adding multiple devices to manually test an issue.  it's not shipped with ZenPack.  self-approving and merging

PR #887, opened 2017-08-25
Remove redundant property `formatted_description` from WinSerice.

mihnatenko:
fixes [ZPS-2042](https://jira.zenoss.com/browse/ZPS-2042).


dbouchillon:
👍

PR #889, opened 2017-08-28
Add app pool status check into iis site datasource

dbouchillon:
Fixes ZPS-2051

Added app pool status check to IISSiteDataSource.  Send events for
status when not in Running state.  send clears for Running. use
datapoint to display current status in grid.  update unit tests to
account for new data being returned.  update readme with changes.

Also consolidated config keys so that we only need to get data once per device instead of per site


dbouchillon:
bump

ZenPacks/zenoss/Microsoft/Windows/datasources/IISSiteDataSource.py
```diff
@@ -183,46 +186,86 @@ def collect(self, config):
...
-        data['values'][ds0.component]['status'] = {
-            'Running': 0,
-            'Stopped': 1}.get(sitestatus, -1), 'N'
+            for result in results[0][results[0].keys()[0]]:
+                site_results[result.Name] = result.ServerAutoStart
+        except Exception:
+            pass
+        app_pool = None
+        app_pool_statuses = {}
+        for line in results[1].stdout:
```
j053ph4: maybe a test here to avoid index errors?

PR #897, opened 2017-09-05
Set zSnmpMonitorIgnore to true for Microsoft device class

dbouchillon:
Fixes ZPS-2108

Assume windows devices don't use snmp.  Set zSnmpMonitorIgnore to true in device_classes.yaml on the /Server/Microsoft class

Also moved device_classes and event_classes to separate yaml files


dougsyer:
I use snmp for hp/dell/ibm agent monitoring.  Id like to not have to fix this on every update

dbouchillon:
Doug, are your hp/dell/ibm snmp devices under the /Server/Microsoft device class?

dougsyer:
We use in house zenpacks for the hp/dell and ibm based agent monitoring but use enterprise zp for ilo/drak monitoring.  I probably should merge our stuff into the enterprise version and talk to the dev but hasnt got to the top of the list.

We have sub device classes organizing by applications and under that server hardware bur we are going to just bind the modelers in zen5 based on the hw product version discovered using a cron job that checks maybe once a week.  We could just flip the snmp bit there but the bigger issue is probably:

We have quite a few windows devices with snmp based device templates for application monitoring.  Even if there are wmi counters we were using snmp if possible because the documentaion in wmi counters is as you know usually non existent and getting counters off is usually a cut/paste effort hoping to use thr nee tools to be able to dump wmi counters available to ext where we can import them

dbouchillon:
We're only going to be setting zSnmpMonitorIgnore on the /Server/Microsoft device class level.  This will only occur on a fresh install.  It will not be set during an upgrade due to how ZPL 2 works.  So, you'll still be able to keep your existing monitoring in place without changing anything.

PR #900, opened 2017-09-06
change default event class to /Status/Winrm

dbouchillon:
Fixes ZPS-1843

Default class should not be /Unknown


dbouchillon:
bump

PR #902, opened 2017-09-07
Possible fix for data gaps

dbouchillon:
Possibly Fixes ZPS-2075

Revert perfmondatasource run time for get-counter to 10 minutes.
This could make a difference so that we restart the command sooner.


eimajenthat:
Approved.

PR #913, opened 2017-09-12
Reformat job results in powershell

dbouchillon:
Fixes ZPS-2143

Not all of the tabs were being printed from the powershell script.
Changed formatting to use 'job_<property>' so that we're not dependent
on some sort of specific characters which may or may not print.

This only seems to affect jobs.  databases and backups not affected.


dbouchillon:
bump

PR #914, opened 2017-09-13
use component for component datasources, contextcompname for device

dbouchillon:
Fixes ZPS-2142

This will allow users to specify ${here/id} tales expression for a component
datasource.


dbouchillon:
bump

PR #917, opened 2017-09-14
Document the need to stop zenjobs during upgrade

dbouchillon:
Fixes ZPS-1853

It also describes how to run a job manually


dbouchillon:
bump...the check failed from a bad unit test that was accidentally merged.  this is only a doc change

dbouchillon:
anyone?  this is only documentation...no code

PR #927, opened 2017-09-26
Fix bound device templates display on the left side pane

dbouchillon:
Fixes ZPS-2187.


dbouchillon:
Cherry-picked commit into hotfix plus update readme

PR:  https://github.com/zenoss/ZenPacks.zenoss.Microsoft.Windows/pull/925

PR #931, opened 2017-10-05
Remove duplicate datasources when Exchange ZenPack is removed

dbouchillon:
Fixes ZPS-2197.


dbouchillon:
reviewed in https://github.com/zenoss/ZenPacks.zenoss.Microsoft.Windows/pull/929

PR #932, opened 2017-10-05
make sure error is string, not list

dbouchillon:
Fixes ZPS-2242

Certain twisted errors return a list of exceptions in a list as the message.
This will stringify them so we can check for kerberos or other specific errors.
Also added comment to OperationTimeout debug message to show that it's expected.


dbouchillon:
bump.  pretty easy one

PR #938, opened 2017-10-13
found better way to allow for unicode apostrophe u'\u2019'

dbouchillon:
Fixes ZPS-2298

This will allow us to always use single quotes around all counters so they
are sent in as literals and will not be expanded because of double quotes.

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/datasources/PerfmonDataSource.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/datasources/PerfmonDataSource.py
@@ -873,11 +873,9 @@ def format_counters(ps_counters):
     """
     counters = []
     for counter in ps_counters:
-        # this will be the best we can do for some foreign language counters
-        if "'" in counter or u'\u2019' in counter:
-            counters.append('(\\"{0}\\")'.format(counter))
-        else:
-            counters.append("('{0}')".format(counter))
+        # check for unicode apostrophe present in foreign langs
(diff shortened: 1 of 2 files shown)


ZenPacks/zenoss/Microsoft/Windows/datasources/PerfmonDataSource.py
```diff
@@ -873,11 +873,11 @@ def format_counters(ps_counters):
...
     for counter in ps_counters:
-        # this will be the best we can do for some foreign language counters
-        if "'" in counter or u'\u2019' in counter:
-            counters.append('(\\"{0}\\")'.format(counter))
-        else:
-            counters.append("('{0}')".format(counter))
+        # check for unicode apostrophe present in foreign langs
+        apostrophe = counter.split(u'\u2019')
+        if len(apostrophe) > 1:
+            counter = "'+[char]8217+'".join(apostrophe)
```
joshw: Is this equivalent to 
counter = counter.replace(u'\u2019', "'+[char]8217+'")
or am i missing something?  (I think i might be, your code won't kick in unless there are two of them in the string, which may be important?)

If your implementation has a reason like that, i approve this change.. it certainly looks functionally fine.

dbouchillon: you know, that does do what i wanted.  not sure why i was using split.

no, the code would kick in if there's 1.  splitting will always return a list with at least 1 item.  i'm changing to replace() though.

PR #939, opened 2017-10-18
Move output to innermost loop to get all disk info

dbouchillon:
Fixes ZPS-2015

the output of each disk object was being overwritten until the very last one.
need to print out each object as we build it


dbouchillon:
bump

PR #940, opened 2017-10-19
send empty disk object maps to nodes to remove duplicates

dbouchillon:
Fixes ZPS-1932

After a failover occurs, modeling will not remove existing disk components for a node


dbouchillon:
bump.  very easy one here

PR #944, opened 2017-10-30
Handle case of empty win32_computersystem

dbouchillon:
Fixes ZPS-2384

This more safely handles empty returns from the OperatingSystem modeler plugin.
One customer is seeing empty win32_computerystem.  this will try to get some of the same info
out of win32_operatingsystem instead and do it safely.  Added unit test both with customer
data and creating empty results


dbouchillon:
bump

dbouchillon:
bump

dbouchillon:
BUMP please

PR #945, opened 2017-10-31
Pull cluster disk state from wmi class

dbouchillon:
Fixes ZPS-2416

Need to be sure to pull the state from the wmi class because we're using
it to determine state, not get-clusterresource.  Also, it's redundant to
model disks in the resources.

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/datasources/ClusterDataSource.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/datasources/ClusterDataSource.py
@@ -130,7 +130,7 @@ class ClusterDataSourcePlugin(PythonDataSourcePlugin):
         cluster_name_items = pipejoin('$_.Name $_.State')
 
         psClusterCommands.append(
-            "get-clusterresource | foreach {{'res-'+{}}};".format(cluster_name_items)
+            "get-clusterresource | where {{ $_.ResourceType.name -ne 'Physical Disk'}} | foreach {{'res-'+{}}};".format(cluster_name_items)
         )
 
         psClusterCommands.append(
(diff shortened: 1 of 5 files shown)


ZenPacks/zenoss/Microsoft/Windows/datasources/ClusterDataSource.py
```diff
@@ -220,7 +207,7 @@ def onSuccess(self, results, config):
                 # specific for cluster disk component
                 data['values'][comp]['state'] = int(state)
                 state = cluster_disk_state_string(int(state))
-            except:
+            except Exception:
                 # handles all other cases
                 data['values'][comp]['state'] = cluster_state_value(state), 'N'
```
pcarinhas: can data['values'][comp]['state'] be both a single value and a tupple?

docs/body.md
```diff
@@ -1819,6 +1819,7 @@ Changes
 
 -   Fix Microsoft Windows fails to model 2008 Server Cluster Disks (ZPS-2015)
 -   Fix Windows - traceback when no data returned for wmi class win32_computersystem (ZPS-2384)
+-   Fix Cluster Disk state always 'Inherited' (ZPS-2416)
```
pcarinhas: Why do we have both a Mediawiki and Markdown? Shouldnt there be only one source?

dbouchillon: b/c windows is special.  it's a free zenpack, so i keep the wiki page up to date.  also, the wiki is where we are linking to download the zenpack.  so, i might as well keep it up for now

ZenPacks/zenoss/Microsoft/Windows/datasources/ClusterDataSource.py
dbouchillon: yes, this is a datapoint and timestamp.  this is how all other datasources are saving datapoints

PR #948, opened 2017-11-03
Fix unit tests.

dbouchillon:
Fixes ZPS-2384

pickle from 5x will not load into 4 due to DateTime difference


dbouchillon:
bump.  pretty simple fix

PR #949, opened 2017-11-03
model unpartitioned cluster disks

dbouchillon:
Fixes ZPS-2437

We should include unpartitioned disks also, not just those with partitions and volumes


dbouchillon:
bump please

dbouchillon:
bump please

PR #951, opened 2017-11-06
Fix "DatapointNameConfict" error occuring in 5x.

dbouchillon:
Fixes ZPS-2421, ZPS-2187

Binding both exchange templates causes a "DatapointNameConfict" (typo in platform) in 5x.  We'll just monkeypatch getTemplates so we can show MSExchange2013IS in the templates tree.


dbouchillon:
bump

ZenPacks/zenoss/Microsoft/Windows/patches/platform.py
```diff
@@ -114,3 +115,30 @@ def _editDetails(self, info, data):
         info.post_update()
 
     return result
+
+
+@monkeypatch('Products.Zuul.facades.devicefacade.DeviceFacade')
+def getTemplates(self, id):
+    object = self._getObject(id)
```
ssvsergeyev: This monkeypatch will affect all devices. Is it intentional or may be a filter by "Windows" only will be better here?

PR #952, opened 2017-11-07
Revert change from previous commit

dbouchillon:
Fixes ZPS-2421

Reverting monkeypatch of platform code as unnecessary after talking to @cluther 

"What I’m thinking is that the platform DeviceFacade.getTemplates should be fixed to only use the `in object.zDeviceTemplates` check if `not isinstance(object, Device)`. I say this because `obj.getRRDTemplates()` has a very different meaning when called on device classes vs. devices.
This way it could be fixed generically, not have to be patched in Windows, and be applicable to other ZenPacks using the same template binding strategy as Windows."


dbouchillon:
bump

dbouchillon:
bump please

PR #954, opened 2017-11-08
Use single eventKey so same error is not duplicated

dbouchillon:
Fixes ZPS-2464

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/modeler/WinRMPlugin.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/modeler/WinRMPlugin.py
@@ -219,7 +219,7 @@ class WinRMPlugin(PythonPlugin):
                 eventClass=eventClass,
                 eventClassKey=key,
                 device=id,
-                eventKey=self.__class__.__name__,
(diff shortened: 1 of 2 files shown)


ZenPacks/zenoss/Microsoft/Windows/modeler/WinRMPlugin.py
```diff
@@ -219,7 +219,7 @@ def _send_event(self, reason, id, severity, force=False,
                 eventClass=eventClass,
                 eventClassKey=key,
                 device=id,
-                eventKey=self.__class__.__name__,
+                eventKey='|'.join(('WinRMPlugin', id)),
```
joshw: This is fine, but couldn't the same thing be accomplished by removing the eventKey altogether?

joshw: I think without an eventKey specified, it constructs a dedupid out of the device/component id + eventClassKey, which ends up being pretty much what you have now.  I might be missing a detail.. anyway, no harm in what you're doing either.

PR #958, opened 2017-11-20
Remodel cluster after owner node changes

dbouchillon:
Fixes ZPS-2134

We should kick off a remodel of a cluster when the owner node changes.
Some of the mapped disks may not exist any longer on a node and can trigger
false alerts.  When modeling a cluster device we'll need to remodel the nodes also.


dbouchillon:
bump

dbouchillon:
bump

ZenPacks/zenoss/Microsoft/Windows/device_classes.yaml
```diff
@@ -90,6 +90,7 @@ device_classes:
       zDeviceTemplates: [Device, Cluster, Active Directory, IIS, MSExchange2010IS]
```
mihnatenko: There are a few files that changed their file mode. Was that planned change, or it happened accidentally?

dbouchillon: yeah, this happens due to where i'm editing my files.  sometimes i edit on my windows, sometimes in linux.  it doesn't affect the finished product

PR #961, opened 2017-11-29
Allow for non-physical adapters to be monitored

dbouchillon:
Fixes ZPS-2567

Added unit test to account for no physical adapters installed but interface
still has a counter.


dbouchillon:
bump

PR #962, opened 2017-11-30
getHostByName will not work in ZaaS, pull from device

dbouchillon:
Fixes ZPS-2645

Should have been using this method in the first place as getHostByName
could hit dns

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/datasources/ShellDataSource.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/datasources/ShellDataSource.py
@@ -20,7 +20,6 @@ import urllib
 from urlparse import urlparse
 from traceback import format_exc
 import re
-from socket import gaierror
 
 from zope.component import adapts
 from zope.component import getGlobalSiteManager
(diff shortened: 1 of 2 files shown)


ZenPacks/zenoss/Microsoft/Windows/datasources/ShellDataSource.py
```diff
@@ -753,8 +751,8 @@ def params(cls, datasource, context):
         if hasattr(context, 'cluster_node_server'):
             owner_node, _ = context.cluster_node_server.split('//')
             try:
-                owner_node_ip = getHostByName(owner_node)
-            except gaierror:
+                owner_node_ip = context.device().clusterhostdevicesdict[owner_node]
+            except Exception:
                 pass
```
pcarinhas: Its more efficient to set the default value to None for owner_node_ip, then do a context.device().clusterhostdevicesdict.get(...) . Having it set to blank string has side effects.

dbouchillon:
thanks Phil, made changes

PR #964, opened 2017-12-05
Fix alias length for iissiteotherrequestmethods__sec to under 30

dbouchillon:
Fixes ZPS-2611

Changed iissiteotherrequestmethods__sec to iissiteotherreqmethods__sec to meet 30 character limit
Also updated analytics bundle with new alias name


dbouchillon:
bump.  very simple one here

PR #966, opened 2017-12-13
get_data should return if datasource is in STOPPED state

dbouchillon:
Fixes ZPS-2717

When we get an immediate error from starting the get-counter command,
we should go ahead and return.  If we do not, this is leaving the ds
in a running state and contributes to missed runs.  looks to be due to the
timing of how long it takes for the error to come back from trying to get data.
The task does eventually finish, but, in the meantime, it appears as a Running_Task
and then eventually a Missed_Run.


dougsyer:
Hurray

PR #969, opened 2017-12-21
Account for event logs with '-' and/or '/' in name

dbouchillon:
Fixes ZPS-2741

Regex was not catching names with symbols, e.g. Microsoft-Windows-Backup or
Microsoft-Windows-PowerShell/Operational


dbouchillon:
bump

PR #973, opened 2018-01-11
No need to remove datapoints/graphs

dbouchillon:
Fixes ZPS-2197

There is no need for this any more.  i think there may have been an issue with the testers system as the counters are valid


dbouchillon:
bump.  very simple one

dbouchillon:
bump

PR #974, opened 2018-01-11
Apply fixes for running tasks issues

dbouchillon:
Fixes ZPS-2829

all windows datasources should have returned from collection before the end of the cycle.
this will enforce it so that nothing goes longer than the cycle time of a datasource.
Also add updated txwinrm.  contains changes that should help this issue and some logging
changes that will make it easier to follow a specific device in the winrm logs

txwinrm 1.3.5 reviewed in https://github.com/zenoss/txwinrm/pull/141

Apply new txwinrm lib changes which include updates for running tasks and memory leaks
Add new z property for specifying a connection timeout.  this will allow users to kill a task faster if necessary
Make the perfmon start command(s) create a DeferredList with individual timeouts.  only try to receive from commands that successfully started
For powershell mssql commands, just increase the operation timeout instead of using a long running command.  may also need to look into revamping the powershell script for pulling sql counters.  this can take quite a while if there are a lot of databases.
fix iis issue found during testing.

txwinrm 1.3.6 reviewed in https://github.com/zenoss/txwinrm/pull/142


dbouchillon:
bump

PR #977, opened 2018-01-29
Remove duplicate exchange dps, graphs, thresholds

dbouchillon:
Fixes ZPS-2197

There are two duplicate datasource/points in the exchange zenpack.  different names for the same counter.  these are overriding the base windows exchange templates


dbouchillon:
bump

PR #979, opened 2018-01-31
Account for IIS scripts/tools not installed on iis7+

dbouchillon:
Fixes ZPS-2728

If user does not install IIS Management Scripts and Tools, we can account for the missing wmi class.  Also add message so that they will be informed to install the required scripts/tools on their device.


ZenPacks/zenoss/Microsoft/Windows/datasources/IISSiteDataSource.py
```diff
@@ -213,10 +213,14 @@ def onSuccess(self, results, config):
...
 
         for ds in config.datasources:
+            sitestatusinfo = None
             try:
                 sitestatusinfo = site_results[ds.params['statusname']]
             except Exception:
-                sitestatusinfo = None
+                try:
+                    sitestatusinfo = site_results[prepId(ds.component)]
+                except Exception:
```
ssOleg: Could we check specific exception type there?

PR #980, opened 2018-02-07
Update txwinrm, includes multiple fixes for memory, tasks

dbouchillon:
Fixes ZPS-2963, ZPS-2176, ZPS-2480, ZPS-2934

should fix issue with memory usage in zenpython.  cpu should be related to the mem fix.  fix during modeling we'll only get one kerberos ticket.  fix for basic auth and long running command.

txwinrm updates were reviewed in https://github.com/zenoss/txwinrm


ZenPacks/zenoss/Microsoft/Windows/datasources/PerfmonDataSource.py
```diff
@@ -231,12 +231,14 @@ def start(self, command_lines):
         create an appropriate set of commands.
         """
         deferreds = []
-        if self.num_commands != len(command_lines):
-            self.commands = self._create_commands(len(command_lines))
+        if self.num_commands != len(self.commands):
+            self.num_commands = len(self.commands)
```
joshw: I can't quite follow this- so the idea here is to re-create self.commands, but only if self.num_commands has changed- it looks like the only way that happens is if remove_corrupt_counters does it?  Is that right so far?

I think i'd like to see a comment here explaining that, if so..

And it looks like self._create_commands sets self.num_commands to the value that was passed in (in this case, self.num_commands's existing value), which is.. confusing.   Because that doesn't really make sense, if it is based on the length of self.commands already..  So this seems like it would all do nothing?    

I think i'm missing something- i don't reall yknow what command_lines is either :)

dbouchillon: well, the only part i changed here was resetting self.num_commands to the new number of commands.  the self._create_commands has always been there.

for the windows perfmon get-counter command, there is a line length of 8192.  if a user has added a new template or created their own with more counters, this could cause the line length to go past that limit and we'd need to create  a new command.  or in the case of the corrupt counters it could lower it so we don't need as many commands.

dbouchillon:
ah, i see what happened.  i pulled that change out as unecessary.  thanks for catching it

PR #984, opened 2018-02-20
Removed default event class for some datasources

dbouchillon:
Fixes ZPS-3056

EventLogDataSource is correct and we will continue to send to the /Unknown event class.
Also updated event datasource section so that we explain the max age field and how it only is used once.


dbouchillon:
bump.  pretty straightforward

PR #989, opened 2018-02-22
Send onError events to '/Status' for datasources

dbouchillon:
Fixes ZPS-3056

Send onError type events to /Status.  no need to eventClassKeys or using datasource eventclass


dbouchillon:
bump

dbouchillon:
bump

dougsyer:
The issue i have here is that we have a ton of things in the status tree and if i need to modify polling events with a transform thrn every status event need to at least go through an if then statement plus any  modifications this high in the event class tree can cause major issues if there is a mistake made in the transform

PR #1011, opened 2018-03-06
clean up releases, show less redundant info

dbouchillon:
Show only last 3 releases.  only need to show that we're compatible with Zenoss.  core and res mgr are implied.  release section was not creating correctly either.  add space between colon and text so that line items show


dbouchillon:
bump...simple doc updates

PR #1012, opened 2018-03-12
Set checked_result to True when looking at db status

dbouchillon:
Fixes ZPS-3194

If a user changes db status datasource to different interval than other db datapoints, we are reporting an error parsing data.  We should only set checked_result if status is sole datasource.

This is in case a user changes the cycle time of the status datasource.  In that case, there should only be the status datasource and we should set checked_result.  If there are other datasources, then we shouldn't set it in case there were actual problems.  updated unit tests with more recent data to include the status datasource

Add note to readme that user shouldn't change cycle time because it adds extra burden on windows device.


dbouchillon:
bump

PR #1014, opened 2018-03-23
Add sql server query optimizations

dbouchillon:
Fixes ZPS-3287

Using SetDefaultInitFields on SMO object to make object loading faster.  https://technet.microsoft.com/en-us/library/ms210395.aspx.  Also when querying counters we only need to run one query.  Write-Host is faster than format-list or format-table.  We can output fewer lines this way.  querying 250 databases now should take under 1 minute.  Updated unit tests to account for new output format of data.


dbouchillon:
bump

ZenPacks/zenoss/Microsoft/Windows/datasources/ShellDataSource.py
```diff
@@ -508,18 +515,22 @@ def parse_result(self, dsconfs, result):
...
+                databasename, _counter, value = db_regex.match(counterline).groups()
+                databasename = databasename.strip()
+                _counter = _counter.strip().lower()
+                value = value.strip()
+            except Exception:
+                log.debug('MSSQL parse_result error in data: %s', counterline)
+                continue
+            if databasename not in self.valuemap:
+                self.valuemap[databasename] = {}
+            if _counter.strip() == 'databasestatus':
```
ssOleg: do we need `_counter.strip()` ? (we have stripped it during creation `_counter = _counter.strip().lower()` )

dbouchillon: thanks for catching that.  i was just leaving some of the original logic which had that.  also found we were double stripping the value as well.

PR #1017, opened 2018-04-09
use sspi in connection string for fewer logins

dbouchillon:
Fixes ZPS-3392

By using SSPI for Windows Authentication, we can avoid passing in the windows user password altogether.  Also it will cut down on the number of logons/logoffs when getting job/db properties.  we'll recommend using windows auth in readme because Microsoft does also.  centralize the connection part of the script because it's used in modeler plugin and shell datasource.


dbouchillon:
bump

PR #1019, opened 2018-04-23
Connect to correct node to pull database, job, backups

dbouchillon:
Fixes ZPS-3461

be sure to connect to the correct device in the conn_info to pull info.  get ip address of node to do this.  Updated unit test for cluster.  query db to get version for clustered sql server.  we should already have version for standalone server so no need to query for it.  add debug log to look at elapsed time for shell datasource


dbouchillon:
bump

ZenPacks/zenoss/Microsoft/Windows/modeler/plugins/zenoss/winrm/WinMSSQL.py
```diff
@@ -84,8 +85,9 @@ def __init__(self, conn_info):
             | Get-ClusterParameter -Name VirtualServerName,InstanceName
             | Group ClusterObject | Select
             @{Name='SQLInstance';Expression={($_.Group | select -expandproperty Value) -join '\\'}},
-            @{Name='OwnerNode';Expression={($ownernode, $domain) -join '.'}}};
-        $cluster_instances | % {write-host \"instances:\"($_).OwnerNode\($_).SQLInstance};
+            @{Name='OwnerNode';Expression={($ownernode, $domain) -join '.'}},
+            @{Name='IPv4';Expression={(Test-Connection $ownernode -count 1).IPV4Address.ipaddresstostring}}};
```
j053ph4: Maybe just do some testing there to see if the IPv4 field is set, and to use IPv6 if it isn’t? and also account for the possibility that neither is set/available?

PR #1020, opened 2018-04-30
fix service status issue, kerb ticket expiry

dbouchillon:
Fixes ZPS-3216

txwinrm reviewed in https://github.com/zenoss/txwinrm/pull/153.  it handles edge case where a ticket expires in the middle of establishing a connection.  also we need to send a clear to /Status.

Additional change for ZPS-3461.  if no ip address returned, then we need to just use the ownernode hostname.


ZenPacks/zenoss/Microsoft/Windows/modeler/plugins/zenoss/winrm/WinMSSQL.py
```diff
@@ -222,8 +222,12 @@ def collect(self, device, log):
                     owner_node = owner_node.strip()
                     sql_server = sql_server.strip()
                     instance = instance.strip()
+                    ip_address = ip_address.strip()
                     conn_info = conn_info._replace(hostname=owner_node)
-                    conn_info = conn_info._replace(ipaddress=ip_address.strip())
+                    if ip_address:
+                        conn_info = conn_info._replace(ipaddress=ip_address.strip())
```
ssOleg: Looks like `ip_address` is already stripped during assigning.

PR #1023, opened 2018-07-23
fix uninitialized local variable

dbouchillon:
Fixes ZPS-3997

errorMessage was uninitialized, causing an errback and self.state was never changed to STOPPED.  state was stuck in the STARTING state and could never be brought out of it due to this.  updated event summary and sending clear events now


dbouchillon:
bump...pretty easy one

PR #1039, opened 2018-08-13
add unit test to ensure output from 2016 servers

dbouchillon:
Fixes ZPS-3902

counters2012 powershell script was not returning any output for 2016 servers previously.  this test ensures that we do have output from a 2016 server

added Interfaces pickle file


dbouchillon:
bump please

PR #1070, opened 2019-03-14
Windows Disconnected Network Drives that Cause PowerShell Error

alexvashchyshyn:
[ZPS-4866](https://jira.zenoss.com/browse/ZPS-4866)

- Handle "Attempting to perform the InitializeDefaultDrives operation on the 'FileSystem' provider failed." error.
- monitoring doesn't stop working


ZenPacks/zenoss/Microsoft/Windows/datasources/PerfmonDataSource.py
```diff
@@ -608,6 +607,15 @@ def _parse_deferred_result(self, result):
                 if stderr:
                     ps_error = ' '.join(stderr)
                     LOG.debug('stderr: {}'.format(ps_error))
+                    if "Attempting to perform the InitializeDefaultDrives operation on the 'FileSystem' provider failed." in ps_error:
```
dbouchillon: actually, this is a red herring and we don't need to report this as an event.  so if you do find this in stderr, then we need to ignore it as the perf data should still be there.  can you also add this check into the search_corrupt_counters method?

this error has nothing to do with actual collection, it's just something that gets written to stderr by powershell when running any script.

ZenPacks/zenoss/Microsoft/Windows/datasources/PerfmonDataSource.py
```diff
@@ -756,6 +764,12 @@ def onReceiveFail(self, failure):
             self.receive()
             defer.returnValue(None)
 
+        elif "Attempting to perform the InitializeDefaultDrives operation on the 'FileSystem' provider failed." in e.message:
```
dbouchillon: because we'll be ignoring the failure, this will be unnecessary

ZenPacks/zenoss/Microsoft/Windows/datasources/PerfmonDataSource.py
```diff
@@ -756,6 +758,12 @@ def onReceiveFail(self, failure):
...
+            retry, level, msg = (
+                True,
+                logging.WARN,
```
dbouchillon: let's change the logging to DEBUG.  and also the message.  "failure" in the message will alarm people and cause support to waste time.  how about "Ignoring powershell error on {} as it does not affect collection: {}"

alexvashchyshyn: great, sure

PR #1071, opened 2019-03-15
allow for cluster sql instance perfmon

dbouchillon:
fixes ZPS-4888

sql server can be active on node other than the current master.  this change will allow for contacting the correct device


dbouchillon:
bump

PR #1076, opened 2019-03-18
update event key so it can be cleared

dbouchillon:
fixes ZPS-4703


dougsyer:
I mean it wont map to the event class the instance is in..

PR #1082, opened 2019-03-22
be sure /Server/Microsoft/Cluster gets zprop assigned

dbouchillon:
Fixes ZPS-4887

clusterOwnerChange only applies to clusters.  make sure the property is set on any z prop under /Server/Microsoft/Cluster including the device class


ZenPacks/zenoss/Microsoft/Windows/migrate/AddClusterOwnerChange.py
```diff
@@ -37,10 +38,13 @@ def migrate(self, dmd):
 
     def update_zprop(self, thing):
         """Ensure clusterOwnerChange is in zWindowsRemodelEventClassKeys."""
+        import pdb; pdb.set_trace()
```
alexvashchyshyn: maybe to get rid of this

PR #1085, opened 2019-03-26
ZPS-5158: fix MSSQL events

alexvashchyshyn:
[ZPS-5158](https://jira.zenoss.com/browse/ZPS-5158)

show /Status/MSSQL eventClass for events connected to MSSQL

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/datasources/ShellDataSource.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/datasources/ShellDataSource.py
@@ -1160,4 +1160,4 @@ def get_valid_dsconf(dsconfs):
     for dsconf in dsconfs:
         if dsconf.eventClass != '':
             return dsconf.eventClass
-    return None
+    return '/Status'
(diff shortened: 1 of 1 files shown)


ZenPacks/zenoss/Microsoft/Windows/datasources/ShellDataSource.py
```diff
@@ -1152,3 +1153,11 @@ def get_script(datasource, context):
...
                   str(datasource.script))
     return script
+
+
+def get_valid_dsconf(dsconfs):
+    '''return a valid eventClass'''
+    for dsconf in dsconfs:
+        if dsconf.eventClass != '':
+            return dsconf.eventClass
+    return None
```
ssOleg: Why do you need `return None` at the end?
One more thing, you have the line `eventClass = dsconf.eventClass or get_eventClass`. It looks like we'll have `eventClass = None` if `get_valid_dsconf` function returns nothing.

alexvashchyshyn: Agree, that should be fixed

PR #1101, opened 2019-04-10
ZPS-5585: fix event class key

alexvashchyshyn:
[ZPS-5585](https://jira.zenoss.com/browse/ZPS-5585)


ZenPacks/zenoss/Microsoft/Windows/datasources/EventLogDataSource.py
```diff
@@ -391,8 +391,8 @@ def onError(self, result, config):
             for ds in config.datasources:
                 data['events'].append({
                     'severity': severity,
-                    'eventKey': 'WindowsEventCollection: {}'.format(ds.params.get('eventid', '')),
-                    'eventClassKey': 'WindowsEventCollection',
+                    'eventKey': 'WindowsEventLogCollection: {}'.format(ds.params.get('eventid', '')),
```
dbouchillon: leave this one as is.  eventKey has to match in order for the event to be cleared.  for an event to be cleared, the device, component, eventClass, and eventKey must match

PR #1104, opened 2019-04-25
check to be sure CommandLine exists, and not None

dbouchillon:
Fixes ZPS-5676

also write unit tests to cover this situation


dbouchillon:
bump

PR #1110, opened 2019-05-10
eliminate duplicate impact relationships

cluther:
Specifying impact relationships under the impacts/impacted_by fields and
the dynamicview_relations field results in Impact having to execute the
relationship-finding code twice. It's only necessary to use
dynamicview_relations to get the relationships into Dynamic View and
into Impact.

Fixes ZPS-5778.

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/tests/test_dynamicview.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/tests/test_dynamicview.py
@@ -1,0 +1,58 @@
+##############################################################################
+#
+# Copyright (C) Zenoss, Inc. 2019, all rights reserved.
+#
+# This content is made available according to terms specified in
+# License.zenoss under the directory where your Zenoss product is installed.
+#
+##############################################################################
+
+from Products.DataCollector.plugins.DataMaps import ObjectMap, RelationshipMap
+
+try:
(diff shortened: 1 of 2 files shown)


ZenPacks/zenoss/Microsoft/Windows/zenpack.yaml
```diff
@@ -131,33 +131,16 @@ classes:
...
-      - all_harddisks
-    impacted_by: []
     dynamicview_views: [service_view]
     dynamicview_relations:
       impacts:
         - all_filesystems
         - all_processes
         - all_ipservices
         - all_cpus
-        - all_interfaces
```
joshw: These are already handled by the platform impact adapters, i assume?

ZenPacks/zenoss/Microsoft/Windows/tests/test_dynamicview.py
```diff
@@ -0,0 +1,58 @@
...
+    from ZenPacks.zenoss.DynamicView.tests import DynamicViewTestCase
+except ImportError:
+    import unittest
+
+    @unittest.skip("tests require DynamicView >= 1.7.0")
+    class DynamicViewTestCase(unittest.TestCase):
+        """TestCase stub if DynamicViewTestCase isn't available."""
+
+
+# TODO: Complete this sometime. Right now, something is better than nothing.
```
joshw: 😬

PR #1129, opened 2020-02-03
ZPS-6861: better handling of "invalid selectors"

ethirolle:
Per ZPS-6861: this PR is for better handling of "invalid selectors" error:
- move it up into "elif '500' in e.message:" block
- set retry=False (this message indicates loss of connection to the device possibly due to reboot or restart of the WinRM service)
This should avoid a retry loop where we make thousands of connections to a single device when we hit the "invalid selectors" 500 error.


ethirolle:
Will attempt to repro issue and test this PR in my lab prior to merging this PR.

ethirolle:
Patch was successful in fixing the ZPS-6861 issue in lab testing (tested back on Feb 5).

PR #1132, opened 2020-12-08
Model SQL Instances with Always On resources

otrach:
Implements ZPS-7366.

* Model SQL Instances with Always On resources on them.
* Add Always On capabilities to WinMSSQL modeller plugin.
* Add zSQLAlwaysOnEnabled property for enabling/disabling Always On functionality.


dbouchillon (approved):
👍

PR #1154, opened 2021-02-23
Collect additional data for some AO components

otrach:
Fixes ZPS-7469.

Collect IDs and relations for Replica and SQL Instance without direct access to SQL Instances.

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/modeler/plugins/zenoss/winrm/WinMSSQL.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/modeler/plugins/zenoss/winrm/WinMSSQL.py
@@ -197,42 +197,76 @@ class SQLCommander(object):
         write-host $results_json;
     '''
 
-    ALL_SQL_INSTANCES_PS_SCRIPT = '''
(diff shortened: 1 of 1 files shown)


ZenPacks/zenoss/Microsoft/Windows/modeler/plugins/zenoss/winrm/WinMSSQL.py
```diff
@@ -197,42 +197,76 @@ def __init__(self, conn_info, log):
...
-            };
-            break;
-        };
-        $result = New-Object 'system.collections.generic.dictionary[string, object]';
-        $result['hostname'] = hostname;
-        $result['instances'] = $instances_list;
-        $result_in_json = ConvertTo-Json $result;
-        write-host $result_in_json
-    '''
+    ALWAYS_ON_RESOURCES_ON_NODE_PS_SCRIPT = (
```
otrach, line 200: Add Replicas and SQL Instances info collection to have this info on node level (before accessing SQL Instance directly as the may be inaccessible).

idubei (approved):

PR #1155, opened 2021-02-24
Set IP addr and SQL ver before instance collection

otrach:
Fixes ZPS-7451.

Set IP address and SQL Instance version before collection on particular SQL Instance.
In case the SQL Instance is inaccessible there will be data about SQL Instance IP address and version.

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/modeler/plugins/zenoss/winrm/WinMSSQL.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/modeler/plugins/zenoss/winrm/WinMSSQL.py
@@ -782,8 +782,6 @@ class WinMSSQL(WinRMPlugin):
             ao_replicas
         )
 
-        log.debug('additional_data[nodes_info] on {} is {}'.format(ag_owner_node_fqdn, additional_data['nodes_info']))
-
         # set IP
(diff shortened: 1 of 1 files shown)


ZenPacks/zenoss/Microsoft/Windows/modeler/plugins/zenoss/winrm/WinMSSQL.py
```diff
@@ -781,14 +782,33 @@ def collect_ao_info_on_node(self, ag_owner_node, connection_info, host_username,
...
         )
 
+        # set IP
+        for sql_instance_name, sql_instance_info in results['sql_instances'].iteritems():
+            if isinstance(sql_instance_info, dict):
+                sql_server_node = sql_instance_info.get('sql_server_node', '')
+                sql_server_node_fqdn = '{}.{}'.format(sql_server_node, ag_owner_node_domain)
+                sql_instance_node_ip = additional_data['nodes_info'].get(sql_server_node_fqdn, {}).get('ip_v4')
+                if sql_instance_node_ip:
+                    results['sql_instances'][sql_instance_name]['sql_host_ip'] = sql_instance_node_ip
```
otrach, line 792: This is DefaultDict

idubei (approved):

PR #1158, opened 2021-03-04
Get proper Default instance original name

otrach:
Implements ZPS-7515.

Default SQL Instances are named after their Windows nodes. Need to take this
into consideration when operating with theirs original names.


ZenPacks/zenoss/Microsoft/Windows/modeler/plugins/zenoss/winrm/WinMSSQL.py
```diff
@@ -651,8 +652,11 @@ def collect_ao_info_on_sql_instance(self, connection_info, instance_hostname, ho
                         if replica_server_hostname and ag_owner_node_domain:
                             replica_hostname_fqdn = '{}.{}'.format(replica_server_hostname, ag_owner_node_domain)
 
-                        instance_title, sqlserver = get_sql_instance_naming_info(instance_name=replica_instance_original_name,
```
otrach, line 654: Also removed sqlserver variable assignment as it may wrongly overrides previously assigned value [here](https://github.com/zenoss/ZenPacks.zenoss.Microsoft.Windows/blob/30445814308aeb4cef23d92e84b49d2a4d821569/ZenPacks/zenoss/Microsoft/Windows/modeler/plugins/zenoss/winrm/WinMSSQL.py#L573) because replica SQL Instance isn't always current SQL Instance.

idubei (approved):

PR #1168, opened 2021-03-17
Adjust value of kerberos_rdns prop to its type

otrach:
Fixes ZPS-5995.

Adjust value of Device's kerberos_rdns property to its type.


ZenPacks/zenoss/Microsoft/Windows/actions.py
```diff
@@ -143,6 +143,8 @@ def _conn_info(self, device):
         code_page = getattr(device, 'zWinRSCodePage', 65001)
         include_dir = getattr(device, 'zWinRMKrb5includedir', None)
         disable_rdns = getattr(device, 'kerberos_rdns', False)
+        if callable(disable_rdns):
```
otrach, line 146: As it is a method: [code](https://github.com/zenoss/ZenPacks.zenoss.Microsoft.Windows/blob/721efc81053a7ccd01850620a88ac10d42d29a61/ZenPacks/zenoss/Microsoft/Windows/BaseDevice.py#L74)

idubei (approved):

PR #1178, opened 2021-05-21
Fix broken monitoring for Perfmon datasoutrce

otrach:
Fixes ZPS-7598.

* Fix bad ComplexLongRunningCommand unique_id as it was not much unique after we had added cluster_node_server uniqueness to Perfmon plugin config_key.
* Add separate timeout for Long Running command for more flexibility.


ZenPacks/zenoss/Microsoft/Windows/datasources/PerfmonDataSource.py
```diff
@@ -961,7 +968,7 @@ def search_corrupt_counters(self, winrs, counter_list, corrupt_list):
             pass
         elif num_counters == 1:
             result = yield add_timeout(winrs.run_command(ps_command, ps_script(counter_list)),
-                                       OPERATION_TIMEOUT + 5)
+                                       self.config.datasources[0].zWinRMConnectTimeout + 5)
```
otrach, line 971: Here we use SingleCommandClient, not Long running one.

ZenPacks/zenoss/Microsoft/Windows/device_classes.yaml
```diff
@@ -2157,7 +2157,7 @@ device_classes:
             strategy: powershell MSSQL AO AG
             component: ${here/id}
             resource: AvailabilityGroupState
-            cycletime: ${here/zWinPerfmonInterval}
+            cycletime: '300'
```
otrach, line 2160: zWinPerfmonInterval need to be used only for Perfmon datasource.

idubei (approved):

PR #1197, opened 2022-02-18
retaining required zProp values

jeff-powell:
Potential fix for ZPS-8015, (if not, then this fixed another undiscovered bug)
The Windows ZenPack tries to be fancy by using 'windows_user', 'windows_password', and 'windows_servername' instead of the zProperties.  however, because these are method-properties of the base object and not zProperties defined in the yaml, zenoss-prodbin, can get really confused and not know what values to put into a proxy device. For CZ, zenmodeler and prodbin's ZenHub/service/ModelerService.py was not providing all the required values to complete a connection.

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/txwinrm_utils.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/txwinrm_utils.py
@@ -46,6 +46,9 @@ ConnectionInfoProperties = (
     'zWinServicesModeled',
     'zWinServicesNotModeled',
     'snmpSysName',
+    'zWinRMUser',
+    'zWinRMPassword',
+    'zWinRMServerName',
 )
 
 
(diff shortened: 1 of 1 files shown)


ZenPacks/zenoss/Microsoft/Windows/txwinrm_utils.py
```diff
@@ -56,96 +59,102 @@ def createConnectionInfo(device_proxy):
...
     auth_type = 'kerberos' if '@' in username else 'basic'
-    if auth_type == 'kerberos' and not device_proxy.zWinKDC:
-        raise UnauthorizedError(
-            "zWinKDC must be configured for domain authentication")
+    if auth_type == 'kerberos' and not winKDC:
+        raise UnauthorizedError("zWinKDC must be configured for domain authentication")
 
-    scheme = device_proxy.zWinScheme.lower()
+    scheme = getProxyValue('zWinScheme')
+    scheme = scheme if hasattr(scheme, 'lower') else ''
```
jpeacock-zenoss, line 96: This is an odd check.  Isn't this just checking whether you have a string?

It also misses the original intent of the code.  The original code was making sure the `scheme` value was lower cased.

ZenPacks/zenoss/Microsoft/Windows/txwinrm_utils.py
```diff
@@ -56,96 +59,102 @@ def createConnectionInfo(device_proxy):
...
-            "zWinRMUser must be user@example.com, not DOMAIN\User")
+    if re.match(r'[a-zA-Z0-9][a-zA-Z0-9.]{0,14}\\[^"/\\\[\]:;|=,+*?<>]{1,104}', username):
+        raise UnauthorizedError("zWinRMUser must be user@example.com, not DOMAIN\User")
 
-    password = device_proxy.windows_password
+    password = getProxyValue(['windows_password', 'zWinRMPassword'])
     if not password:
-        raise UnauthorizedError(
-            "zWinRMPassword or zWinPassword must be configured")
+        raise UnauthorizedError("zWinRMPassword or zWinPassword must be configured")
```
jpeacock-zenoss, line 88: The message mentions "zWinPassword", but that zprop isn't searched for in the preceeding getProxyValue() call.

ZenPacks/zenoss/Microsoft/Windows/txwinrm_utils.py
```diff
@@ -56,96 +59,102 @@ def createConnectionInfo(device_proxy):
...
+                val = attr() if callable(attr) else attr
+                if val:
+                    return val
+        return ""
+
+    hostname = getProxyValue(['windows_servername', 'zWinRMServerName', 'manageIp'])
+    if not hostname:
+        raise UnauthorizedError("Attempted Windows connection to non-Windows device")
+
+    hostname = getProxyValue(['windows_servername', 'zWinRMServerName', 'manageIp'])
```
sbourg-zenoss, line 77: appears to be a duplicate line reassigning the value from line 73

ZenPacks/zenoss/Microsoft/Windows/txwinrm_utils.py
```diff
@@ -56,96 +59,102 @@ def createConnectionInfo(device_proxy):
...
+        raise UnauthorizedError("zWinScheme must be either 'http' or 'https'")
+
+    try:
+        port = int(getProxyValue('zWinRMPort'))
+        if port not in (5986, 443) and scheme == 'https':
+            raise Exception()
+        if port not in (5985, 80) and scheme == 'http':
+            raise Exception()
+    except Exception:
         raise UnauthorizedError("zWinRMPort must be 5986 or 443 if zWinScheme is https")
```
sbourg-zenoss, line 109: Line 106 checks for http and port (80, 443) but error message does not state http configuration is a valid setting.

jpeacock-zenoss (changes requested):

ZenPacks/zenoss/Microsoft/Windows/txwinrm_utils.py
jeff-powell: You're right, zWinPassword was taken out many versions ago.  However, the message was intentionally kept this way (for reasons lost to time.)  I don't see why it needs to be here anymore, removing.

ZenPacks/zenoss/Microsoft/Windows/txwinrm_utils.py
jeff-powell: This is my paranoia of a typeless language haunting me, lol. But getProxyValue(x) will always return a string or None object.  updated.

ZenPacks/zenoss/Microsoft/Windows/txwinrm_utils.py
jeff-powell: correct.  I extended the error message

jpeacock-zenoss (approved):

PR #1198, opened 2022-02-22
Fix MSSQL Database addition and replacement templates

oherlinger:
Handle ZPL returning addition and replacement templates for WinDatabase and WinAODatabase


odubrovyk:
It has been fixed in https://github.com/zenoss/ZenPacks.zenoss.Microsoft.Windows/pull/1239

PR #1201, opened 2022-04-07
Add zProperty for Windows services class grouping

odubrovyk:
Fixes ZPS-8115.

Added zWinServicesGroupedByClass property for Windows Service
classes to model services with generic Windows Service class.

### DIFF
--- a/ZenPacks/zenoss/Microsoft/Windows/modeler/plugins/zenoss/winrm/Services.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/modeler/plugins/zenoss/winrm/Services.py
@@ -115,7 +115,5 @@ def truncate_service_class(group_list, service_name, service_caption):
             if re.search(group, service_name):
                 service_class_name = re.search(group, service_name).group(0)
                 service_class_caption = re.sub("(?:.(?!_.*))+$", '', service_caption)
-                return service_class_name, service_class_caption
-        return service_class_name, service_class_caption
-    else:
-        return service_class_name, service_class_caption
+                break
+    return service_class_name, service_class_caption
--- a/ZenPacks/zenoss/Microsoft/Windows/tests/testServicesRegex.py
+++ b/ZenPacks/zenoss/Microsoft/Windows/tests/testServicesRegex.py
@@ -126,10 +126,21 @@ class TestServicesRegex(BaseTestCase):
         self.assertEqual(service_class_name, 'PimIndexMaintenanceSvc_aeff1234')
         self.assertEqual(service_class_caption, 'Contact Data_aeff1234')
 
+    def test_model_service_grouped_multiple_underscores(self):
+        self.device.zWinServicesNotModeled = []
+        self.device.zWinServicesModeled = []
+        self.device.zWinServicesGroupedByClass = ['Test_Windows_Service']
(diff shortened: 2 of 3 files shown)


StevePC (changes requested):
Need some documentation explaining what the contents of the zProp should be and exactly what it does.

docs/body.md
```diff
@@ -1292,6 +1292,11 @@ important.
 - zWinDBSnapshotIgnore
     :   MS SQL Database Snapshots modeling disabled. Set to true to disable modeling and monitoring of MS SQL Database snapshots.
 
+- zWinServicesGroupedByClass
+    :   List of regular expressions for Windows services to model with generic Windows Service class. Specify class names to set generic Windows Service class names for matching Windows Services.
+        This property should be used when a target Windows device has a lot of Windows Services with rotating name extension - "<service_name>_0123456789bcdef". Each of these services creates separate Windows Service class which could cause ZODB lock.
```
StevePC, line 1297: probably don't need to mention ZODB locking, but should mention that the regex is used in search() not match(), so not anchored to the start of the name, and that the regex supplied has no bearing on what is removed from the service name to arrive at the service class name, it simply removes the last underscore and everything following it.

ZenPacks/zenoss/Microsoft/Windows/tests/testServicesRegex.py
```diff
@@ -96,3 +107,52 @@ def test_model_service_multiply_values(self):
...
+    def test_model_service_grouped_empty(self):
+        self.device.zWinServicesNotModeled = []
+        self.device.zWinServicesModeled = []
+        self.device.zWinServicesGroupedByClass = []
+        _, _, group_list = validate_modeling_regex(self.device, self.log)
+        service_class_name, service_class_caption = truncate_service_class(group_list, 'PimIndexMaintenanceSvc_aeff1234', 'Contact Data_aeff1234')
+        self.assertEqual(service_class_name, 'PimIndexMaintenanceSvc_aeff1234')
+        self.assertEqual(service_class_caption, 'Contact Data_aeff1234')
+
+    def test_model_service_grouped_multiple(self):
```
StevePC, line 129: might want to include tests for grouping services with multiple underscores in the name

StevePC (changes requested):

ZenPacks/zenoss/Microsoft/Windows/modeler/plugins/zenoss/winrm/Services.py
```diff
@@ -103,3 +105,17 @@ def omit_service(model_list, ignore_list, service_name, log):
...
+    service_class_caption = service_caption
+    if group_list:
+        for group in group_list:
+            if re.search(group, service_name):
+                service_class_name = re.search(group, service_name).group(0)
+                service_class_caption = re.sub("(?:.(?!_.*))+$", '', service_caption)
+                return service_class_name, service_class_caption
+        return service_class_name, service_class_caption
+    else:
+        return service_class_name, service_class_caption
```
jeff-powell, line 121: I  don't think you need the 'else' section.  Just set the default to return at the end.

jeff-powell (changes requested):
Made some suggestions, but looks great.  StevePC also dropped in a comment too (regarding the documentation.)

StevePC (approved):

PR #1205, opened 2023-02-16
Fix failed perfmon collection for Windows hosts

odubrovyk:
Fixes ZPS-8015.

Added WinRM port check before LongRunning command execution. 
This prevents PerfMon datasource from hanging when one of the hosts 
with Kerberos authentication stops receiving WinRM connections. 
Added correct default values to counters dict to prevent possible attribute errors. 
Added appropriate unit test.


otrach (approved):
As an workaround it is a good solution.

PR #1213, opened 2023-07-12
Add WinRM port check timeout property

odubrovyk:
Fixes ZPS-8015.

Windows ZP side changes related to ZPS-8622.
The new zProperty zWinRMPortCheckTimeout with a default value of 1 second has been added. It controls the timeout for Windows device WinRM Port Check. The default value has been set for 1 second according to the average netcat command runtime on customers' environments.


ZenPacks/zenoss/Microsoft/Windows/__init__.py
```diff
@@ -144,6 +144,9 @@ class ZenPack(schema.ZenPack):
                                                               'label': 'Windows remodel event class keys'},
                             'zWinRMConnectTimeout': {'type': 'int',
                                                      'description': 'Used to define the time out for a winrm connection.'},
+                            'zWinRMPortCheckTimeout': {'type': 'float',
```
vsaliieva, line 147: will be good to add default value here

vsaliieva (changes requested):
add notes about new zproperty to documentation

vsaliieva (approved):

PR #1239, opened 2025-02-10
Add replacement/addition templates support

odubrovyk:
Fixes ZPS-8798.

Add replacement/addition templates support for
WinService and WinSQLDatabase models.


ZenPacks/zenoss/Microsoft/Windows/WinService.py
```diff
@@ -36,7 +36,16 @@ def getRRDTemplateName(self):
         return 'WinService'
 
     def getRRDTemplates(self):
-        return [self.getRRDTemplateByName(self.getRRDTemplateName())]
+        template_name = self.getRRDTemplateByName(self.getRRDTemplateName())
```
StevePC, line 39: this should probably be just getRRDTemplateName() instead of getRRDTemplateByName(getRRDTemplateName())

StevePC (changes requested):

StevePC (approved):

PR #1242, opened 2025-02-25
Fix perfmon collection if the PS command is too long

odubrovyk:
Fixes ZPS-8643.

We divide the full command string length on the counters_limit variable to determine the number of commands to run.
Then assuming that all counters have a similar length we divide the list of counters into <num_commands + 1> parts. The length of some counters is bigger than the others and the total length for those commands is bigger than 8192 chars. To ensure that the length of each command is lower than the limit the data source recursively finds the optimal number of commands.


ZenPacks/zenoss/Microsoft/Windows/datasources/PerfmonDataSource.py
```diff
@@ -422,22 +422,34 @@ def _build_commandlines(self):
             self.num_commands = 0
             return
 
-        # The number of long running commands needed to get data for all
+        def normalize_chunks(chunks, num_commands):
+            if all(len(format_counters(cgroup)) < counters_limit for cgroup in
+                   chunks):
+                return chunks, num_commands
+            else:
+                num_commands += 1
```
StevePC, line 430: won't this make the line longer each recursion?  I think it should be num_commands -= 1

StevePC (changes requested):

StevePC (approved):

